PR #83, opened 2015-05-13
Remove edit/delete endPoint - it make no sense in the context of simplelayout

### LINKED ISSUE #73: Use url given from server for block actions

maethu:
It's in the context of a block, and we shop the url with the component actions (addable blocks)
@bierik 

Closes #73


bierik:
@maethu pls use double-quotes

maethu:
@bierik  done

PR #84, opened 2015-05-13
Components are added at the end.

bierik:
Connect jQuery's draggable to the sortable lists for adding components
anywhere.
Place toolbox at the bottom and make it collapsable.

Adding a layout when start dragging a block is still in progress.
See SO question: http://stackoverflow.com/questions/30195800/create-new-sortable-container-on-drag-start


maethu:
As long as SO question is open, we're not able to fix this issue properly. 

Workaround:
OnLayoutDelete: Disable Components if the last layout is removed. 
OnLayoutAdd: Enable Components

PR #88, opened 2015-05-13
Make simplelayout configurable through data attributes.

### LINKED ISSUE #87: Use data attributes for configuration
Make possible to use ```data-settings="{option: value}"``` for overriding the default settings on simplelayout container.

bierik:
Fixes #87 

Extend default config with specific settings.

### DIFF
--- a/ftw/simplelayout/browser/resources/main.js
+++ b/ftw/simplelayout/browser/resources/main.js
@@ -13,6 +13,8 @@
       simplelayout: null,
       init: function(callback) {
         var self = this;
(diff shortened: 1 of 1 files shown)


ftw/simplelayout/browser/resources/main.js
```diff
@@ -13,6 +13,8 @@
       simplelayout: null,
       init: function(callback) {
         var self = this;
+        var settings = $(this.settings.source).data("settings") || {};
```
maethu: What about sl-settings?

PR #121, opened 2015-06-08
ftw.contentpage to ftw.simplelayout migration

maethu:
:construction: 

- Migration based on plone.app.contentttypes migrations.
- Migrate ftw.contentpage ContentPage to ftw.simplelayout.ContentPage
- Migrate ftw.contentpage TextBlock to ftw.simplelayout.TextBlock
- Sample Migration of Imageblock (Image treated as block)

This is not a full migration of all content, currently ftw.simplelayout does not provide all features of ftw.contentpage, so there are several missing steps, like contentcategories field migration, or adressblock migration, etc.


maethu:
New migration is on the way.

PR #152, opened 2015-06-24
Remove unused font definition.

bierik:
Fixes 4teamwork/bern.web#157

Definition is done by integration.
Update client lib css.


bierik:
@maethu updated

PR #168, opened 2015-07-03
Uploading images in filelistingblock will create a file instead an image

elioschmutz:
@maethu Uploading images in filelistingblock will create a file instead an image

closes: https://github.com/4teamwork/bern.web/issues/183

### DIFF
--- a/ftw/simplelayout/tests/test_uploadcapable.py
+++ b/ftw/simplelayout/tests/test_uploadcapable.py
@@ -1,0 +1,38 @@
+from collective.quickupload.interfaces import IQuickUploadFileFactory
+from ftw.builder import Builder
+from ftw.builder import create
+from ftw.simplelayout.browser import uploadcapable
+from ftw.simplelayout.testing import FTW_SIMPLELAYOUT_CONTENT_TESTING
+from Products.ATContentTypes.interfaces.file import IATFile
+from unittest2 import TestCase
+from zope.component import queryAdapter
+
+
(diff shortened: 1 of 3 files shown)


ftw/simplelayout/tests/test_uploadcapable.py
```diff
@@ -0,0 +1,38 @@
...
+    def test_adapter_registration_on_listingblock(self):
+        adapter = queryAdapter(self.filelistingblock, IQuickUploadFileFactory)
+        self.assertIsNotNone(adapter)
+        self.assertIsInstance(
+            adapter, uploadcapable.FileListingQuickUploadCapableFileFactory)
+
+    def test_portal_type_is_alway_a_file(self):
+        adapter = queryAdapter(self.filelistingblock, IQuickUploadFileFactory)
+        # creates
+        adapter('test.jpg', None, None, None, None, None)
```
maethu: @elioschmutz Please upload a real file, otherwise this line is not clear what it does.

ftw/simplelayout/tests/test_uploadcapable.py
```diff
@@ -0,0 +1,38 @@
...
+        contents = self.filelistingblock.listFolderContents()
+
+        self.assertNotEqual(
```
maethu: better assertion would be: self.assertEquals(1, len(contents), 'Expect one item').
This way the second assertion is needless.

maethu:
Just check the two assertions, everything else looks fine!! :sparkles:

PR #174, opened 2015-07-09
Do not update page config if a block is deleted because of an ancesto…

buchi:
…r of him is deleted.

This allows us to delete Plone sites again :smile:


maethu:
aww my bad... thanks for fixing it :relieved:

PR #182, opened 2015-07-15
Synchronize loading of requirejs library.

bierik:
Fixes https://github.com/4teamwork/bern.web/issues/218

According to 4teamwork/bern.web#218 an
```undefined is not a function ``` error occured randomly. This was
because of the asynchronous loading of the simplelayout library even in
the compiled version. Thanks to
https://github.com/sahat/requirejs-library two fragments are required
for wrapping the compiled file. These fragments synchronises the loading
process.

/cc @buchi


bierik:
@jone @buchi updated

PR #186, opened 2015-07-21
Raises exception if block has no component definition.

bierik:
Fixes https://github.com/4teamwork/bern.web/issues/238
When no component is defined make a warning.


maethu:
imho this does not fix the source of the issue.

We had some Issue wit Constrain types in plone, which prevents simplelayout form getting all addable blocks. 

Since we removed this Behavior, the error no longer appears.

PR #189, opened 2015-07-22
Remove spaces and newlines between tags in structure.pt - This allows…

maethu:
… us to use "empty" selector in css.

It's implement with your own ViewPageTemplateFile to avoid clashes
with evaluated content. I now removes spaces and newlines before the
pt is evaluated.


maethu:
cc @buchi

jone:
How does IE handle this? 
I remember that we used to have problems with ``<div class="visualClear"></div>`` in IE because that tag was dropped because it had no content. Could this lead into trouble with IE?

PR #196, opened 2015-07-27
Add scss factory, which hides all block types from the add menu.

maethu:
Hide block types from add menu with ftw.theming and without ftw.theming. 

:exclamation: the `!important` is necessary, since the `authoring.css` from plone also has a !important.

### DIFF
--- a/ftw/simplelayout/browser/dynamic_scss_resources.py
+++ b/ftw/simplelayout/browser/dynamic_scss_resources.py
@@ -1,0 +1,29 @@
+from ftw.simplelayout.utils import get_block_types
+from ftw.simplelayout.utils import normalize_portal_type
+from ftw.theming.interfaces import ISCSSResourceFactory
+from ftw.theming.resource import SCSSResource
+from zope.interface import provider
+
+
+TEMPLATE = """
+#plone-contentmenu-factories {{
+  {0} {{
+    display: none !important;
+  }}
+}}
+"""
+
+
+@provider(ISCSSResourceFactory)
+def hide_blocks_in_factory_menu(context, request):
+
+    selectors = ['']
+
+    for block_fti in get_block_types():
+        selectors.append(
+            u'.contenttype-{0}'.format(
+                normalize_portal_type(block_fti.getId())))
+
+    return SCSSResource('simplelayout_hide_blocks.scss',
+                        slot='addon',
+                        source=TEMPLATE.format(', '.join(selectors)))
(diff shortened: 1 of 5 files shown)


ftw/simplelayout/utils.py
```diff
@@ -14,3 +19,17 @@ def get_block_html(block):
...
     return block.restrictedTraverse(viewname)()
+
+
+def get_block_types():
+    platform = getSite()
+    types_tool = getToolByName(platform, 'portal_types')
+    typeids = types_tool.objectIds()
+    typeids.sort()
+    for type_id in typeids:
+        dx_fti = queryUtility(IDexterityFTI, name=type_id)
```
jone: Testing the actual FTI for `plone.dexterity.interfaces.IDexterityFTI` would be more efficient, since we would not have to make a utility lookup for each FTI each time.
Something like this should work:
```python
block_ftis = filter(IDexterityFTI.providedBy, types_tool.objectValues())
return map(methodcaller('getId'), block_ftis)
```

ftw/simplelayout/browser/dynamic_scss_resources.py
```diff
@@ -0,0 +1,29 @@
...
+    display: none !important;
+  }}
+}}
+"""
+
+
+@provider(ISCSSResourceFactory)
+def hide_blocks_in_factory_menu(context, request):
+
+    selectors = ['']
```
jone: Why the empty string?

ftw/simplelayout/utils.py
maethu: Yep, you're right, I'll change that.

ftw/simplelayout/browser/dynamic_scss_resources.py
maethu: faulty left over :smile: - will be removed.

maethu:
@jone updated

ftw/simplelayout/utils.py
```diff
@@ -14,3 +19,14 @@ def get_block_html(block):
...
+
+
+def get_block_types():
+    platform = getSite()
+    types_tool = getToolByName(platform, 'portal_types')
+
+    dexterity_ftis = filter(
+        IDexterityFTI.providedBy, types_tool.objectValues())
+
+    return filter(
```
jone: This returns the FTI-object, but I think it should return the ID of the FTI, right=

jone:
:+1:

ftw/simplelayout/utils.py
maethu: The [addable_blocks_view](https://github.com/4teamwork/ftw.simplelayout/blob/master/ftw/simplelayout/browser/ajax/addable_blocks_view.py#L30) expects ftis

PR #199, opened 2015-07-28
Dragging layout is very buggy and may cause a site crash.

bierik:
Disable shrinking the layouts when dragging for preventing side effects
which may cause a site crash. Use tolerance intersects for preventing
flickering when the layouts changes their position.


maethu:
Please remove commented code, if we want this back we search for this commit.

bierik:
@maethu updated

PR #201, opened 2015-07-28
Move mapblock into its own buildout extra and Plone profile.

mbaechtold:
This way the mapblock is no longer installed by default and won't pull in its heavy dependencies.

Closes 4teamwork/bern.web#281


jone:
@mbaechtold Since the mapblock now is a separate feature, I'd prefer to have it as a separate package which contains all the relevant parts (content type schema and class, view, generic setup profile, etc).
Then we could just load the ZCML with a condition in `ftw.simplelayout/configure.zcml`, e.g.:
```xml
<include package=".mapblock" zcml:condition="installed collective.geo.bundle" />
```
What do you think?

mbaechtold:
@jone, I like that idea. @maethu, is this ok for you?

maethu:
:+1: for a separate package

mbaechtold:
I have moved the map block stuff into its own package.

ftw/simplelayout/mapblock/permissions.zcml
```diff
@@ -0,0 +1,10 @@
+<configure
+    xmlns="http://namespaces.zope.org/zope"
+    i18n_domain="ftw.topics">
```
jone: I think this is the wrong domain :wink:

ftw/simplelayout/mapblock/browser/configure.zcml
```diff
@@ -0,0 +1,20 @@
+<configure
+    xmlns="http://namespaces.zope.org/zope"
+    xmlns:browser="http://namespaces.zope.org/browser"
+    xmlns:zcml="http://namespaces.zope.org/zcml"
+    i18n_domain="ftw.simplelayout">
+
+    <include package=".viewlets" />
+
+    <configure zcml:condition="installed collective.geo.mapwidget">
```
jone: `collective.geo.mapwidget` is included in [`collective.geo.bundle`](https://pypi.python.org/pypi/collective.geo.bundle).
Since this file is only loaded when `collective.geo.bundle` is installed, we can assume that `collective.geo.mapwidget` and don't need the `zcml:condition`.

ftw/simplelayout/configure.zcml
```diff
@@ -59,12 +65,12 @@
...
-      name="development"
-      title="ftw.simplelayout development"
-      directory="profiles/development"
-      description='Javascirpt resources from client are not compressed'
-      provides="Products.GenericSetup.interfaces.EXTENSION"
-      />
+        name="development"
+        title="ftw.simplelayout development"
+        directory="profiles/development"
+        description='Javascript resources from client are not compressed'
```
jone: Can you change this to double quotes in favor of local consistency?

ftw/simplelayout/mapblock/browser/configure.zcml
```diff
@@ -0,0 +1,25 @@
...
+    i18n_domain="ftw.simplelayout">
+
+    <browser:resourceDirectory
+        name="ftw.simplelayout.mapblock.resources"
+        directory="resources"
+        />
+
+    <include package=".viewlets" />
+
+    <configure zcml:condition="installed collective.geo.mapwidget">
```
jone: :scissors:

ftw/simplelayout/mapblock/contents/configure.zcml
```diff
@@ -0,0 +1,22 @@
+<configure
+    xmlns="http://namespaces.zope.org/zope"
+    xmlns:five="http://namespaces.zope.org/five"
+    xmlns:genericsetup="http://namespaces.zope.org/genericsetup"
+    xmlns:i18n="http://namespaces.zope.org/i18n"
+    xmlns:zcml="http://namespaces.zope.org/zcml"
+    i18n_domain="ftw.simplelayout">
+
+    <!-- XXX Implement a specific KMLLayer for Simplelayout and do not recycle the one from geo.kml -->
```
jone: @maethu why should this be specific?

ftw/simplelayout/mapblock/contents/configure.zcml
```diff
@@ -0,0 +1,22 @@
...
+    <configure zcml:condition="installed collective.geo.kml">
```
jone: :scissors: `collective.geo.kml` is available because of `collective.geo.bundle` dependency.

ftw/simplelayout/mapblock/contents/configure.zcml
mbaechtold: I don't know. This has been there before my refactoring.

ftw/simplelayout/mapblock/profiles/default/types/ftw.simplelayout.MapBlock.xml
```diff
@@ -0,0 +1,68 @@
+<?xml version="1.0"?>
+<object name="ftw.simplelayout.MapBlock"
+        meta_type="Dexterity FTI"
+        xmlns:i18n="http://xml.zope.org/namespaces/i18n"
+        i18n:domain="ftw.simplelayout" >
+
+    <!-- Basic metadata -->
+    <property name="title" i18n:translate="">MapBlock</property>
+    <property name="description" i18n:translate="">Use this block to display a map</property>
```
jone: Could you remove this description?
I like the FTI description to always be empty because we often dont want any description and if there is a value here it is impossible to remove it just by using translations.

ftw/simplelayout/profiles/mapblock/metadata.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<metadata>
+  <version>1000</version>
+  <dependencies>
+    <dependency>profile-ftw.simplelayout:default</dependency>
```
jone: I think this profile is no longer necessary, since you have moved it to the `mapblock` package, right?

ftw/simplelayout/mapblock/profiles/default/types/ftw.simplelayout.MapBlock.xml
mbaechtold: Do I need to remove `i18n:translate=""` too?

mbaechtold: And what about the description of the other blocks (in the parent package)? Shall I remove the description too?

ftw/simplelayout/profiles/mapblock/metadata.xml
mbaechtold: True. This is a rebase failure.

jone:
:+1:

ftw/simplelayout/mapblock/permissions.zcml
mbaechtold: True. I have changed this.

ftw/simplelayout/mapblock/browser/configure.zcml
mbaechtold: Done.

ftw/simplelayout/configure.zcml
mbaechtold: Done.

ftw/simplelayout/mapblock/browser/configure.zcml
mbaechtold: Done.

ftw/simplelayout/mapblock/contents/configure.zcml
mbaechtold: Done.

ftw/simplelayout/mapblock/contents/configure.zcml
maethu: There's a adapter lookup somewhere in c.geo for a specific *geo* View marker interface. All geo viewlets are assigned to this view interface. BUT we don't want those viewlets on a block. So I had to define a separate adapter for block views.

PR #203, opened 2015-07-30
Implement leadimage

maethu:
- Also modify the parent on block edit.


ftw/simplelayout/browser/leadimage.py
```diff
@@ -0,0 +1,50 @@
...
+from plone.app.uuid.utils import uuidToObject
+from plone.memoize import ram
+from Products.Five.browser import BrowserView
+import json
+import re
+import hashlib
+
+
+def _render_cachkey(method, self):
+    key = (str(self.context.modified().millis()),
```
jone: Since the cachekey is global, it should contain a distinct identifier.
I would add the method dottedname to the cachekey.

ftw/simplelayout/handlers.py
```diff
@@ -1,9 +1,13 @@
 from Acquisition import aq_parent
+from DateTime import DateTime
 from ftw.simplelayout.interfaces import IPageConfiguration
 from ftw.simplelayout.interfaces import ISimplelayout
 from persistent.list import PersistentList
 from persistent.mapping import PersistentMapping
 from plone.uuid.interfaces import IUUID
+from Products.CMFPlone.interfaces.siteroot import IPloneSiteRoot
+from zope.event import notify
+from zope.lifecycleevent.interfaces import IObjectModifiedEvent
```
jone: :scissors:

ftw/simplelayout/handlers.py
```diff
@@ -70,3 +74,18 @@ def update_page_state_on_block_remove(block, event):
                         # Block has been removed
                         break
         config.store(page_state)
+
+
+def modify_parent_on_block_edit(block, event):
+    parent = aq_parent(block)
```
jone: should `aq_parent` always be used in combination with `aq_inner` :grey_question:
```python
parent = aq_parent(aq_inner(block))
```

ftw/simplelayout/tests/builders.py
```diff
@@ -11,6 +13,12 @@ class ContenPageBuilder(DexterityBuilder):
 class TextBlockBuilder(DexterityBuilder):
     portal_type = 'ftw.simplelayout.TextBlock'
 
+    def with_dummy_image(self):
+        image = Path(__file__).joinpath('..', 'assets', 'fullhd.jpg').abspath()
```
jone: Is it relevant that the image is this big? A smaller image might speed up the tests a little :wink: 
How about `mario.gif`?

ftw/simplelayout/tests/test_leadimage.py
```diff
@@ -0,0 +1,106 @@
...
+from ftw.simplelayout.testing import FTW_SIMPLELAYOUT_CONTENT_TESTING
+from ftw.simplelayout.testing import IS_PLONE_5
+from ftw.simplelayout.testing import SimplelayoutTestCase
+from ftw.testbrowser import browsing
+from plone.uuid.interfaces import IUUID
+from unittest2 import skipUnless
+import transaction
+
+
+@skipUnless(not IS_PLONE_5, 'requires plone < 5')
```
jone: How about?
```python
@skipIf(IS_PLONE_5, ....)
```
I would understand it too, then :grin:

ftw/simplelayout/tests/test_parent_modified_event.py
```diff
@@ -0,0 +1,40 @@
...
+        cached_date = page.modification_date
+
+        block = create(Builder('sample block').within(page))
+
+        browser.login().visit(block, view='edit')
+        browser.find_button_by_label('Save').click()
+        self.assertNotEquals(cached_date,
+                             page.modification_date,
+                             'Modification date of sl container should change')
+
```
jone: can you add a test where the block changes instead of adding a new block?

maethu: caching the date after adding new block, makes much more sense, right ;-)

ftw/simplelayout/tests/builders.py
maethu: Well, I need an image with a huge dimension... ok at least with a 400px width. I chose an image, which is large enough for all plone default scales.

ftw/simplelayout/handlers.py
maethu: don't know, at least it doesn't hurt.

ftw/simplelayout/tests/test_parent_modified_event.py
jone: Ah well, that's why I was confused.
Adding does change the container anyway..

ftw/simplelayout/tests/builders.py
jone: :+1:

PR #216, opened 2015-08-17
Remove behaviors, which are not necessary

maethu:
As discussed with @jone ftw.simplelayout should only provide basic behaviors.
This means the decision, which dexterity metadata behaviors should be active on a contentpage should delegated to the integration.


jone:
:smile:

PR #222, opened 2015-08-19
Implement redirect to parent view only for anonymous users (otherwise…

maethu:
… show folder_contents).

The new View is applied to the FileListingBlock and the GalleryBlock.

### DIFF
--- a/ftw/simplelayout/browser/configure.zcml
+++ b/ftw/simplelayout/browser/configure.zcml
@@ -37,6 +37,13 @@
         />
 
     <browser:page
+        for="plone.folder.interfaces.IFolder"
+        name="redirect_to_parent"
+        class=".redirect_to_parent.BlockContainerRedirectToParent"
+        permission="zope2.View"
+        />
+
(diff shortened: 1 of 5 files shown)


ftw/simplelayout/browser/configure.zcml
```diff
@@ -37,6 +37,13 @@
         />
 
     <browser:page
+        for="*"
+        name="bock_container_redirect_to_parent"
```
jone: Typo `bock`.

Wouldn't it be easier to also name the view `redirect_to_parent` but register it for `plone.folder.interfaces.IFolder`? This would automatically use this view for all folderish content, so there is only one right name to put in the FTI XML. But do we want that?
Just an idea...

maethu: that's sounds good, the naming was an issue :smile: this way it's gone!

maethu: @jone updated.

PR #231, opened 2015-08-24
move default contenttypes to "contenttypes" package.

jone:
:construction: Includes #230

    In order to cleanup up extras dependencies conditions, we move all
    things related to default content types into the subpackage
    ftw.simplelayout.contenttypes so that we can only include this package
    when the corresponding extras was installed.
    This makes it easier to not break Plone 4 installations without the
    contenttypes extras installed, because we have a clear separation for
    lib things and contenttype things.


jone:
```
error while rendering ftw.simplelyout.openlayers.support.for.blocks
```
test missing?

maethu:
@jone The Viewlet render bug is fixed in https://github.com/4teamwork/ftw.simplelayout/pull/232

maethu:
cc @buchi This should have "probably" no impact on plone 5 sites :bomb:

PR #233, opened 2015-08-25
Add css class to broken links after editing a block

elioschmutz:
@bierik Could you review my js?

- Add css class to broken links after editing a block
- Add css class to broken links after reloading a page if the logged-in user has edit permission

### DIFF
--- a/ftw/simplelayout/browser/resources/main.js
+++ b/ftw/simplelayout/browser/resources/main.js
@@ -7,7 +7,7 @@
       return global["xhr_" + $(".main-uploader").attr("id")]._filesInProgress > 0;
     };
     var markBrokenLinks = function(blocks) {
-      $("a.internal-link[href*='resolveuid']", $(blocks)).addClass("broken-link");
+      $("a.internal-link[href*='resolveuid']", blocks).addClass("broken-link");
     };
     var initializeColorbox = function() {
       if($(".colorboxLink").length > 0) {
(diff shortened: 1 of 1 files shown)


ftw/simplelayout/browser/resources/main.js
```diff
@@ -6,7 +6,9 @@
     var isUploading = function() {
       return global["xhr_" + $(".main-uploader").attr("id")]._filesInProgress > 0;
     };
-
+    var markBrokenLinks = function(blocks) {
+      $('a.internal-link[href*="resolveuid"]', $(blocks)).addClass("broken-link");
```
bierik: It seems that ``` blocks ``` are already DOM or jQuery elements otherwise you can name it ``` selector ```. So according to http://api.jquery.com/jquery/ there no need to wrap a DOM or jQuery element again for using it as context for a jQuery selector. So theres no need to wrap ``` blocks ``` again.

ftw/simplelayout/browser/resources/main.js
```diff
@@ -106,6 +108,10 @@
         $(simplelayout.options.source).sortable("disable");
       }
 
+      if (instance.settings.canEdit) {
+        markBrokenLinks($('.sl-block'));
```
bierik: Please use double quotes.

elioschmutz:
@bierik :+1:

jone:
Is this only marking links *within* blocks?
We probably also have to mark links *outside* of blocks, which would mean the javascript would have to be moved to a more general place (theme, policy, ...)

bierik:
When it's possible to add content without the simplelayout I would say too that the javascript has to get moved. In that case I think it's feature for the bear.

jone:
If it would be in the theme, we'd have to be able to make it customizable, I guess.
We'll need to implement it in the policy and register to the simplelayout events.

PR #235, opened 2015-08-25
Mapblock searchresults not working.

### LINKED ISSUE #229: MapBlock: Geo-Search triggers "overlay close"
Currently it's not possible to search for a place in the mapblock.

bierik:
Fixes #229 

Mapblock has triggered an overlay submit event instead of asynchronously
loading the location suggestions. The overlay's load method does not
instantly show the overlay. So if the speed is set to 0 make the overlay
visible manually before loading the overlay.

### DIFF
--- a/ftw/simplelayout/browser/resources/Overlay.js
+++ b/ftw/simplelayout/browser/resources/Overlay.js
@@ -46,6 +46,9 @@
       onCancelCallback: function(){},
       onSubmitCallback: function(){},
       open: function() {
+        if(this.settings.speed === 0) {
+          this.element.show();
+        }
         this.overlay.load();
         this.bindCloseEvents();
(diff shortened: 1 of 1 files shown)


maethu:
Well, would it be better to load the geo js stuff after the overlay is ready?

bierik:
According to https://github.com/4teamwork/ftw.simplelayout/blob/master/ftw/simplelayout/mapblock/browser/resources/mapblock.js#L12 the mapblock is loaded after the overlay is ready. But this is still to early. No idea why this event is not working properly.

bierik:
Fixed through https://github.com/OneGov/plonetheme.onegovbear/pull/53

bierik:
@maethu updated, sorry for the confusion

Issue #221, opened 2015-08-18
Translate hardcoded action labels.


mbaechtold:
Examples: "Move this layout around", "Delete this layout"

jone:
Also please change the englisch version of `Move this layout around`, sounds strange. `Move this layout` is sufficient.

bierik:
Fixed through https://github.com/4teamwork/ftw.simplelayout/pull/236

PR #236, opened 2015-08-25
Redesign for toolbox

bierik:
Make redesign for toolbox according to 
https://feedback.onegovbox.ch/t/platzierung-der-toolbox-im-neuen-simplelayout/29/10

bildschirmfoto 2015-08-25 um 16 02 49


bierik:
@jone rebased

jone:
There are no icons when simplelayout is installed without ftw.theming. What's the idea there?
bildschirmfoto 2015-08-26 um 14 13 53

ftw/simplelayout/browser/client/dist/simplelayout.css
```diff
@@ -445,50 +439,81 @@
...
+  margin-right: 15px;
+  background-color: #2c2c34;
+  padding: 4px 7px;
+  color: #fff;
+  border-radius: 2px;
+  float: right;
+  margin-top: -2px;
+  visibility: hidden;
+  opacity: 0;
+  transition: opacity .1s ease-in, margin-right .1s ease-out;
```
jone: I'd remove that. It feels much better without any transition. My two cents. :wink:

jone:
We really should add labels for "blocks", "layouts" and for each layout...
But we might want to add those along with fixing JS translations -> https://github.com/4teamwork/ftw.simplelayout/issues/221

bierik:
@jone I have no idea what is to do with the icons. Whats on your mind concerning ftw.theming. How do we handle updates generally for packages that have not installed ftw.theming? For example if we want to updated ftw.simplelayout where ftw.theming is not installed.

bierik:
@jone As for the animation. I like the animation. Imagine you add 10 blocks in 270 days. While trying to search the right block in the toolbox you hover 20 times over each label equals 100ms * 20 * 10 * 270 equals 1.5 hours a year you have to wait. Not considering that the author is getting faster every day.

maethu:
Ions:
How about going the plone way.

- By default ftw.simplelayout is using the icon configured on the FTI (incl. icon as images).
- The Integration may set the ICON EXPR on the FTI to 'empty' and you need to implement the icon on your own.

jone:
@bierik regarding the **animation**: it is not wait time in this case, I just perceive it as disturbing when quickly hovering over all the icons. I think it is much more pleasant when there is no animation. I see absolutely no benefit in having an animation there. It has no UI nor a UX purpose here. It tells me nothing and it does not make it look nicer, so I just think it is not necessary. :wink: 

**Icons:** variants
- stick with the default Plone FTI icon machinery
- add some default icons and register them with CSS as background
- or: how about just putting the first character of the translated type name there? e.g. `T` for textblock? This would fit with the layout numbering in terms of font / text, right?

maethu:
@jone 

Currently I'm for the easiest way to implement. Obviously using the first letter of the typename may clash with other Blocks. Of course also a solvable Problem (Like using two letters on a name clash. 

Another point:
Users will click on the blocks and the layouts, what should happen in this case?
imho a popup should appear with some simple instructions. 

I also would prefer to handle all translatable label/texts in python code, no JS. 
We have the addable_blocks_view, which is in use to render the addable blocks in the toolbox. 

I would recommend to rename this view into `toolbox_view`, or something similar and put all the necessary data into this json:

- Addable blocks  + Block Actions
- Layout Actions
- Translations

bierik:
I also prefer the easiest way for the icon labels (first two letter of the name) to not threaten the schooling.
Because we're all three available tomorrow it should be possible to move the layout actions and translations to the backend IMO.

@jone you're right I'm going to get rid of the animations.

PR #245, opened 2015-09-03
Add fix for firefox outline bug

elioschmutz:
Fixes https://github.com/4teamwork/bern.web/issues/583

Use ``` box-shadow ``` instead of ``` outline ``` for indicating drop areas.


bierik:
@elioschmutz test passed

PR #251, opened 2015-09-07
Fix layout issues

bierik:
Fixes: https://github.com/4teamwork/bern.web/issues/620

 - Display layouts in toolbox based on permission
 - Update layoutactions after block has been moved


maethu:
:sparkling_heart:

PR #262, opened 2015-09-30
Page get downloaded several times on reload through XHR.

bierik:
The ```blockInserted``` event was binded before the simplelayout DOM
structure got serialized. So every time when a block was
inserted a new Overlay was loaded. Because the URL for getting the form
was empty the whole page got downloaded.

Sorry for that :disappointed:


maethu:
@bierik 
:money_with_wings: :arrow_right: :pig2:

maethu:
I do not really understand what exactly is/was the issue. I just can say on my local machine the issue seems to be solved.

PR #264, opened 2015-10-01
Implement spinner for overlays.

bierik:
bildschirmfoto 2015-10-01 um 11 53 19


ftw/simplelayout/browser/resources/integration.default.css
```diff
@@ -24,6 +24,18 @@
   display: none;
 }
 
+.sl-spinner {
+  $spinner-size: 50px;
```
jone: What's that? Where is this used?

maethu:
I thought we said, that we DO NOT implement a overlay spinner in simplelayout, instead we use the one from plone. 

Because we can nod handle every usecase with the overlay spinner, for example adding new content, obviously can not use the overlay spinner, since the overlay is not yet here. To mix it up with the ajax spinner from plone may cause strange visuals. 

So we decided to use the spinner and it's implementation from the installed theme.

bierik:
@maethu correct this will be fixed through https://github.com/OneGov/plonetheme.onegovbear/pull/69

PR #265, opened 2015-10-01
Disable ajax spinner for toolbox.

bierik:
Depends on https://github.com/OneGov/plonetheme.onegovbear/pull/69

### DIFF
--- a/ftw/simplelayout/browser/resources/main.js
+++ b/ftw/simplelayout/browser/resources/main.js
@@ -25,7 +25,7 @@
 
     var instance = {
       settings: {
-        toolboxDataEndpoint: baseUrl + "sl-toolbox-view",
+        toolboxDataEndpoint: baseUrl + "sl-toolbox-view" + "?spinner=false",
         saveStateEndpoint: baseUrl + "sl-ajax-save-state-view",
         source: ".sl-simplelayout",
         layouts: [1, 2, 4],


jone:
You wrote `depends on`, is this a problem when `ftw.simplelayout` is installed without `plonetheme.onegovbear`?
simplelayout should not have a dependency on any theme, we should make sure that it works with any theme :wink:

bierik:
I just set the spinner flag here for disabling the spinner in case the ```plonetheme.onegovbear``` is active. So where would you implement this functionality though. IMO hiding graphical elements such as the spinner should be done in a theme and not in a third party package like ```ftw.simplelayout```. In https://github.com/OneGov/plonetheme.onegovbear/pull/69 Im gonna unbind an eventhandler so I think it's not a good idea to do this generally.

By the way installing the ```plonetheme.onegov``` without ```ftw.simplelayout``` produces the following error:
```
assError: Error parsing block:

{
display: none !important;
}


on line 5 of dynamic:simplelayout_hide_blocks.scss
Traceback:
  File "/Users/kbieri/.buildout/eggs/pyScss-1.3.4-py2.7-macosx-10.9-x86_64.egg/scss/compiler.py", line 374, in manage_children
    self._manage_children_impl(rule, scope)
  File "/Users/kbieri/.buildout/eggs/pyScss-1.3.4-py2.7-macosx-10.9-x86_64.egg/scss/compiler.py", line 385, in _manage_children_impl
    block = UnparsedBlock(rule, c_lineno, c_property, c_codestr)
  File "/Users/kbieri/.buildout/eggs/pyScss-1.3.4-py2.7-macosx-10.9-x86_64.egg/scss/rule.py", line 342, in __init__
    self.header = BlockHeader.parse(prop, has_contents=bool(unparsed_contents))
  File "/Users/kbieri/.buildout/eggs/pyScss-1.3.4-py2.7-macosx-10.9-x86_64.egg/scss/rule.py", line 258, in parse
    elif prop.split(None, 1)[0].endswith(':'):
IndexError: list index out of range
```

So ```plonetheme.onegovbear``` has a dependency on ```ftw.simplelayout``` :wink:

PR #268, opened 2015-10-12
Change the "store" page configuration behavior

maethu:
This makes it possible to have multiple simplelayout views with different slots on one object. 

Example:
1. Default simplelayout view (only one slot -> `default`
2. Special search view with one slot -> `search`

The old behavior always reassigned the new store, which means if someone edited the simplelayout page with the `search` slot, the informations of the `default` slot was doomed.

The new implementation does now an `update` instead. This way only the sots used in the DOM are updated. 

For this new usecase I also hat to change the behavior what happen if a block was not present in store. 
**Before:** Get all columns and append the block to the first column. 

**New:** Get the first column of the `default` slot.

### DIFF
--- a/ftw/simplelayout/configuration.py
+++ b/ftw/simplelayout/configuration.py
@@ -44,12 +44,17 @@ def make_resursive_persistent(conf):
     return persist(conf)
 
 
-def columns_in_config(config):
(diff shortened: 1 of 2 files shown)


jone:
:+1:

PR #271, opened 2015-10-26
Allow initialize new slots and only check the layout state not the whole storage.

maethu:


ftw/simplelayout/configuration.py
```diff
@@ -148,21 +154,24 @@ def load(self):
...
-
-            if flatten_old != flatten_new:
+        if api.user.has_permission('ftw.simplelayout: Change Layouts',
+                                   obj=self.context):
+            return
+
+        # The user does not have the permission to manipulate the
+        # structure. But the new configuration is serialized from the DOM
+        # So we need to check if the user has manipulated the DOM by
+        # himself. This is done by checking the actual and the new state
```
jone: `actual` -> `stored` ?

ftw/simplelayout/configuration.py
```diff
@@ -148,21 +154,24 @@ def load(self):
...
+        # without the block informations. Initialize a new state is possible,
+        # since the may added thru a template change.
```
jone: ?? somethings wrong with this sentence,

maethu:
@jone updated.

PR #272, opened 2015-10-29
Es 902 accessibility

elioschmutz:
@maethu 

- Remove title attribute on galleryblock image
- Remove title attribute on textblock image and teaser url

### DIFF
--- a/ftw/simplelayout/contenttypes/browser/templates/textblock.pt
+++ b/ftw/simplelayout/contenttypes/browser/templates/textblock.pt
@@ -6,8 +6,7 @@
   i18n:domain="ftw.simplelayout">
 
   <h2 tal:condition="view/block_title">
-      <a tal:attributes="href teaser_url;
-                         title view/block_title"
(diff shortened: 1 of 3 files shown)


ftw/simplelayout/contenttypes/browser/templates/textblock.pt
```diff
@@ -16,7 +15,6 @@
     <div tal:attributes="class image/wrapper_css_classes">
         <a tal:omit-tag="not: image/link_url"
            tal:attributes="href image/link_url;
-                           title image/link_title;
```
jone: Can you remove that from the python code too?
https://github.com/4teamwork/ftw.simplelayout/blob/9cbb2a3d3c211dd11f283bdab5f312584b09fc0f/ftw/simplelayout/contenttypes/browser/textblock.py#L44

elioschmutz: :+1:

PR #278, opened 2015-11-24
Fix unwrap_persistance handler method

elioschmutz:
Fix an issue if the page_config is only partial persistent after retrieving a versioned contentpage (I don't know why it is partial persistent, but it is). Serializing a persistent dict or map is not possible and we get a json encode error.

If the ``unwrap_persistance`` method gets a partial persistent config, it will stop unwrap.

This PR will fix this. The unwrapper will unwrapp all possible elements, doesn't matter if it is already unwrapped or not.

### DIFF
--- a/ftw/simplelayout/handlers.py
+++ b/ftw/simplelayout/handlers.py
@@ -14,11 +14,11 @@ def unwrap_persistence(conf):
     """Unwrap recursice persistent page state
     """
     def unwrap(data):
-        if isinstance(data, PersistentMapping):
+        if isinstance(data, (PersistentMapping, dict)):
             data = dict(data)
             for key, value in data.items():
                 data[key] = unwrap(value)
-        elif isinstance(data, PersistentList):
+        elif isinstance(data, (PersistentList, list, tuple, set)):
             return list(map(unwrap, data))
         else:
             # Usually we got basestrings, or integer here, so do nothing.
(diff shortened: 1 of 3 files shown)


ftw/simplelayout/handlers.py
```diff
@@ -14,11 +14,11 @@ def unwrap_persistence(conf):
     """Unwrap recursice persistent page state
     """
     def unwrap(data):
-        if isinstance(data, PersistentMapping):
+        if isinstance(data, PersistentMapping) or isinstance(data, dict):
```
jone: you can combine those
```
if isinstance(data, (PersistentMapping, dict)):
```

ftw/simplelayout/handlers.py
```diff
@@ -14,11 +14,11 @@ def unwrap_persistence(conf):
...
             data = dict(data)
             for key, value in data.items():
                 data[key] = unwrap(value)
-        elif isinstance(data, PersistentList):
+        elif isinstance(data, PersistentList) or isinstance(data, list):
```
jone: Please also include `tuple` and `set` here
- we always want to recurse (it might contain a dict, right?)
- we want to convert tuples and sets to lists too, since json knows only lists
```python
if isinstance(data, (PersistentList, list, tuple, set)):
```

elioschmutz:
@jone :+1:

Issue #276, opened 2015-11-17
Drag'n'drop behavior


jone:
The drag'n'drop behavior when moving blocks is strange.
It depends on the dragging vector but it should depend on the cursor position.

We should switch the default jquery ui dragging behavior and not configure it.

PR #280, opened 2015-11-24
Improve drag'n'drop behaviour.

bierik:
Fixes https://github.com/4teamwork/ftw.simplelayout/issues/276

``` intersects ``` does not exist and pointer does not satisfy the drag behaviour. The standard setting makes it best.


jone:
Wohoo :sparkles: thanks

PR #281, opened 2015-11-26
Make videoblock.js available for anonymous users too

elioschmutz:
The videoblock is broken for anonymous users because the videoblock.js is not available for anonymous users.

This PR will fix this.

@maethu do we need already upgradesteps?


maethu:
Since we are in Prod mode on several instance, you're invited to create the first upgrade step :smile:

elioschmutz:
:+1: upgradestep created

elioschmutz:
remember...

PR #293, opened 2016-01-13
Add Simplelayout view as possible front page layout + addable types on the plone site

maethu:
No uprade step:
On a customer installation we already had to configure this, otherwise simplelayout did not work properly, that's why I provide no upgrade step for this change


maethu:
@jone updated.

PR #294, opened 2016-01-15
Use default table listing for the listingblock styles.

maethu:
- Implement some sane default withs for columns.
- Do not sort the table with js
screen shot 2016-01-15 at 09 19 33

### DIFF
--- a/ftw/simplelayout/browser/resources/integration.theme.scss
+++ b/ftw/simplelayout/browser/resources/integration.theme.scss
@@ -46,7 +46,7 @@
   opacity: 0;
 }
 
-
+// Galleryblock
 .galleryblockImageWrapper{
   display: table-cell;
   float: left;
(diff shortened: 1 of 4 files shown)


maethu:
cc @jone

jone:
:+1:

ftw/simplelayout/browser/resources/integration.theme.scss
```diff
@@ -75,3 +75,45 @@
...
+
+        &.column-getContentType {
+          width: 16px;
+        }
+
+        &.column-modified {
+          width: 85px;
+        }
+
+        &.column-portal_type {
```
bierik: In the screenshots I can see a column named `type`. But `column-getContentType` or `column-portal_type` suggest a column for displaying a portal type. So which of these classes refer to the screenshot and what about the class which is not in the screenshot.

ftw/simplelayout/browser/resources/integration.theme.scss
```diff
@@ -75,3 +75,41 @@
...
+          width: 25px;
+          padding-right: 0;
+        }
+
```
maethu: @bierik 
Nice catch, it seems like the "column-getContentType" was removed and replaced with "column-portal_type". I remove this.

column-modified => Bearbeitet
column-getObjSize => Grösse
column-portal_type => Typ

All other columns do not matters, since they don't have any specific styles. 

FYI:
`portal_type` usually means `File`, `Link`, `ContentPage`, etc. `content_type` means `PDF`, `XLSX`, etc.

Issue #303, opened 2016-02-03
Simplelayout edit bar - display issue


maethu:
<img width="413" alt="screen shot 2016-02-03 at 18 16 07" src="https://cloud.githubusercontent.com/assets/437933/12790386/5190fc04-caa2-11e5-9709-d32eafafd42f.png">

PR #304, opened 2016-02-09
The editbar button are not floating next to each other when the block is too small.

bierik:
Fixes https://github.com/4teamwork/ftw.simplelayout/issues/303#event-537880922

bildschirmfoto 2016-02-09 um 10 36 48


bierik:
@jone updated

bierik:
@maethu updated

PR #306, opened 2016-02-12
Respect disable_border key in request and do not load simplelayout if present.

maethu:
On `REQUEST.set('disable_border', '1')` the simplelayout toolset is not loaded. 
This is a default plone feature. 

This also affects the simplelayout portlet:
If the portlet is empty and disable_border=1 do not show the portlet.

### DIFF
--- a/ftw/simplelayout/portlets/portlet.py
+++ b/ftw/simplelayout/portlets/portlet.py
@@ -30,8 +30,9 @@ class Renderer(base.Renderer):
         if self.has_blocks():
             return True
         else:
+            disabled = int(self.request.get('disable_border', 0)) == 1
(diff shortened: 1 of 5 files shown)


jone:
:+1:

maethu:
@jone updated

ftw/simplelayout/portlets/portlet.py
```diff
@@ -30,8 +30,9 @@ def available(self):
         if self.has_blocks():
             return True
         else:
+            disabled = self.request.get('disable_border', 0) == '1'
```
jone: I'd make a boolean check: `disable_border` may be `"1"`, `1`, `True` or more..

maethu:
@jone updated

PR #307, opened 2016-02-15
Remove leadimage cache key.

jone:
The cachekey currently includes the full URL to the image.
This is problematic since the URL may change depending how the site is
accessed (VirtualHostMonster, port tunneling).

The second problem is that the cache key does not include the page. This
may lead to the wrong image beeing displayed when multiple pages are
modified at the exact same time.

Since this is cache is probably a premature optimization I'm just
removing it. We can implement a new one when we are in need.


maethu:
:+1:

PR #310, opened 2016-02-19
Implement image shadow action on textblocks

bierik:
⚠️ 🚧 Diskussion in #16 
Fixes https://github.com/4teamwork/plonetheme.blueberry/issues/16

bildschirmfoto 2016-02-19 um 11 33 29
bildschirmfoto 2016-02-19 um 11 33 36


maethu:
@bierik This feature may be added to a specific website... always, so all images has the same styling.

PR #311, opened 2016-02-29
Configure TinyMCE resource types.

jone:
Configure TinyMCE so that the page and the blocks are linkable in TinyMCE and the folderish types (page, listing and gallery blocks) can be opened for linking their contents.


mbaechtold:
Cool, thank you @jone!

PR #317, opened 2016-04-08
Some blocks can be hidden now.

mbaechtold:
Hidden blocks will not be shown on its simplelayout container if the user does not have the permission to edit content.

Users allowed to edit the block will see the block anyway (highlighted).

This is not a security feature, the block and its content can still be accessed if the path is known.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,7 +5,9 @@ Changelog
 1.2.1 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Some blocks can be hidden now. Users allowed to edit the block will see
+  the block anyway.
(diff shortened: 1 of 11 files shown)


jone:
Can you do a rebase in order to fix the tests?

I've realized that you've just hidden the content with CSS (`display: none`), but my idea was to actually not render the HTML when it is hidden.
What was your thoughts? Not sure which is better.

mbaechtold:
@jone, I have rebased onto the `master` branch. Not rendering the block is better. I'll update the PR accordingly.

mbaechtold:
@jone, I have updated this pull request. Hidden block are no longer rendered at all for users not having the permission to edit content.

PR #331, opened 2016-05-03
Implement opengraph support

maethu:
- New viewlet for Opengraph meta tags
- On PloneRoot (Enable/Disable in the control panel)
  - og:title --> Title of Website
  - og:type --> by default 'website', can be configured thru the control panel
  - og:image --> Logo of Website
  - og:url --> URL of Website
- On Simplelayout pages (enable/disable thru a behavior)
  - og:site_name --> Shows the title of the website (page belongs to this site)


jone:
On the site root I see no opengraph tags in the HTML. Can you look into that again?

maethu:
Well the OG tags are available on siteroot, but not on the plone default page, which is a `Document`.

I guess implementing the OG tags for `Document` is not necessary here (even the wrong place)
1. Simplelayout has no default pages and the plone root neither. 
2. I you have simplelayout, you don't use the Document anymore.

maethu:
cc @jone

PR #339, opened 2016-05-25
Implement request tracker.

bierik:
The request tracker tracks URLs which are not banned
through a white and blacklist system. When a request lasts
longer than a specific threshold (300ms by default) the root element
will be extended with a marker class (ajax-loading)

### DIFF
--- a/ftw/simplelayout/browser/resources/RequestTracker.js
+++ b/ftw/simplelayout/browser/resources/RequestTracker.js
@@ -1,0 +1,116 @@
+(function(global, $) {
+
+  "use strict";
+
+  var URLWhiteList = [
+    "sl-ajax-reload-block-view",
+    "sl-toolbox-view",
+    "sl-ajax-delete-blocks-view",
+    "add_block",
+    "sl-ajax-edit-block-view",
+    "sl-ajax-upload-block-view",
+    "edit.json",
+    "tiny_mce_gzip.js"
+  ];
+
+  var URLBlackList = [
+    "@@z3cform_validate_field"
+  ];
+
+  function getParameterByName(name, url) {
+    name = name.replace(/[\[]/, "\\[").replace(/[\]]/, "\\]");
+    var regex = new RegExp("[\\?&]" + name + "=([^&#]*)");
+    var results = regex.exec(url);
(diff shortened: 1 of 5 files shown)


jone:
changelog 😉

ftw/simplelayout/browser/resources/RequestTracker.js
```diff
@@ -0,0 +1,107 @@
...
+      if(checkURL(url)) {
+        requests[url] = Request(url, mark);
+      }
+    }
+
+    function untrack(url) {
+      if(requests[url]) {
+        requests[url].cancel();
+        delete requests[url];
+        unmark();
```
jone: When having concurrent requests, the first finished request will unmark and the spinner will disappear, even when more tracked requests are pending.
This is an "academical" issue, since the user cannot click when the first request begins and thus the user can not trigger more requests.
But depending on the implementation, more requests may be triggered automatically.
If we can fix it easily and without refactoring everything, I would like to do it.

Proposal:
```diff
     function untrack(url) {
       if(requests[url]) {
         requests[url].cancel();
         delete requests[url];
-        unmark();
       }
+      if (Object.keys(requests).length === 0) {
+        unmark();
+      }
     }
```
The idea is that we only remove the spinner / unmark when all tracked requests are finished.
@bierik What do you think?
⚠️ `Object.keys(requests).length` may not be the best solution; I'm no JavaScript expert.

ftw/simplelayout/contenttypes/profiles/default/jsregistry.xml
```diff
@@ -8,4 +8,11 @@
         insert-after="++resource++ftw.simplelayout.resources/main.js"
         />
 
+    <javascript
+        id="++resource++ftw.simplelayout.resources/RequestTracker.js"
```
jone: What is the reason you placed that in the `contenttypes` package?
I think that this is useful regardless of whether the contenttypes are installed.
I would move it to the `lib` profile.

ftw/simplelayout/browser/resources/RequestTracker.js
bierik: Done in https://github.com/4teamwork/ftw.simplelayout/pull/339/files#diff-cae21d3d0441539589f4d26ed6afaf85R79

PR #345, opened 2016-06-17
Add min-height to empty Simplelyout portlet.

maethu:
This fixes a issue, when it was no longer possible to re-add a layout after deleting the last one.


bierik:
👍

PR #350, opened 2016-07-14
Use zindex system from ftw.theming

bierik:
Also fixes iframeblock boundaries:

Before:
bildschirmfoto 2016-07-14 um 17 23 57

After:
bildschirmfoto 2016-07-14 um 17 24 05


jone:
@bierik please rebase

bierik:
@jone updated

PR #361, opened 2016-08-15
Support migration from ftw.contentpage.

jone:
The `migration.py` provides a set of inplace migrators for migrating AT
types (`ftw.contentpage`) to DX types (`ftw.simplelayout`).

The migration must be executed in an upgrade step in the integration
package / project, because:

- a migration upgrade step can't be put in neither the old
  package (`ftw.contentpage`) nor the new package (`ftw.simplelayout`),
  because neither can be sure that the other package is present at a
  specific upgrade step (version)

- in a project the types may have had additional fields bevore (schema
  extender) and / or after (additional behaviors)

- such migrations should be tested well and implemented carefully for a
  specific project

Because of those reasons, we provide pre-configured inplace migrators
for common simplelayout types for various packages here, but they must
be executed in an upgrade step in the project code.

The inplace migrators cover the common problems when migrating
simplelayout types.

- ContentPageMigrator
- NewsFolderMigrator
- NewsMigrator
- EventFolderMigrator
- EventPageMigrator
- ImageToSliderPaneMigrator
- TextBlockMigrator
- AddressBlockMigrator
- HTMLBlockMigrator
- ListingBlockMigrator

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -384,13 +384,23 @@ You can disable (Simplelayout Settings - Control Panel) Opengraph for the plone 
 On Simplelayout sites itself the OpenGraph meta tags can be controlled by the `OpenGraph marker behavior`.
 
 
+Migration from ftw.contentpage
+==============================
+
+This package is the successor of
+`ftw.contentpage <https://github.com/4teamwork/ftw.contentpage>`_.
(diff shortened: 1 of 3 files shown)


jone:
@maethu @mbaechtold I think this one is ready.

PR #364, opened 2016-08-16
Fix image overlay in textblock

bierik:
The colorbox does not get initialized when adding
a textblock with an image.


maethu:
@bierik pls add chagelog

bierik:
@maethu updated

PR #367, opened 2016-08-19
fix icon with image description

Tschanzt:
@maethu I added an additional div which allows me to place the icon relative to the image, since it only contains the image instead of the complete content with caption.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 1.8.2 (unreleased)
 ------------------
 
+- Fix Icon position. It is now postitioned to the image not the container.
+  [tschanzt]
+
 - Add optional title to the video block (disabled by default).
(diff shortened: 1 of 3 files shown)


maethu:
@Tschanzt 
Ist the additional wrapper really necessary?

And:
- Please add changelog
- Indent html if the wrapper is really necessary.

Tschanzt:
@maethu yes, it is necessary, since we need a position:relative element to position our absolute element to. currently the next one is the wrapping div, which contains both the image and the caption and therefore our icon will be moved when the height of the caption changes. So i added a new wrapper which is positioned relative and wraps nothing other than our image.

mbaechtold:
@Tschanzt, could you please add a changelog entry, as requested by @maethu. That would be awesome.

Tschanzt:
@mbaechtold updated.

PR #368, opened 2016-08-24
Fix drop animation

bierik:
The drop animation changed the background color of the dropped
element to `#fff`. The background-color should not be done using
style attribute because overriding these styles requires an `!important`
statement in the CSS.

drop_animation


bierik:
@maethu rebased

mbaechtold:
There is no animation in Firefox, @bierik 💭

mbaechtold:
👍

PR #379, opened 2016-09-05
Add show_title checkbox for contentpage

raphael-s:
Adds a checkbox for contentpages which allows a user to decide whether the title of the contentpage is displayed or not.
This works with the new `IContentPageShowTitle` behavior.

See: https://github.com/4teamwork/winterthur.web/issues/312

### DIFF
--- a/ftw/simplelayout/browser/simplelayout.py
+++ b/ftw/simplelayout/browser/simplelayout.py
@@ -7,7 +7,8 @@ from zope.interface import implements
 from zope.publisher.browser import BrowserView
 import json
 import logging
-
+from ftw.simplelayout.interfaces import IContentPageShowTitle
+from zope.component import queryAdapter
 LOG = logging.getLogger('ftw.simplelayout')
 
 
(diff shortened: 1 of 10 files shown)


ftw/simplelayout/contenttypes/profiles/default/types/ftw.simplelayout.ContentPage.xml
```diff
@@ -33,6 +33,7 @@
     <property name="behaviors">
         <element value="ftw.simplelayout.interfaces.ISimplelayout" />
         <element value="ftw.simplelayout.opengraph.interfaces.IOpenGraphSupport" />
+        <element value="ftw.simplelayout.interfaces.IContentPageShowTitle" />
```
jone: @raphael-s We don't need this behavior by default. When needed, we can activate it in the customer project.

So I propose to:

- [ ] Do not enable it in `ftw.simplelayout.ContentPage.xml`
- [ ] Make sure the changes in `simplelayout.pt` do not break when the behavior is not available
- [ ] Document the existence of the behavior in the Readme.

raphael-s:
@Tschanzt added a few tests, I think this should be ready for a merge now :)

ftw/simplelayout/browser/simplelayout.py
```diff
@@ -7,7 +7,8 @@
 from zope.publisher.browser import BrowserView
 import json
 import logging
-
+from ftw.simplelayout.interfaces import IContentPageShowTitle
+from zope.component import queryAdapter
```
Tschanzt: ✂️

Tschanzt:
👍

PR #384, opened 2016-09-14
Protect new show/hide title behavior.

maethu:
I thought calling the permission this way makes more sense, since we are protecting it from getting hidden :-)

### DIFF
--- a/ftw/simplelayout/tests/test_contentpage_behavior.py
+++ b/ftw/simplelayout/tests/test_contentpage_behavior.py
@@ -8,7 +8,7 @@ from zope.component import getUtility
 import transaction
 
 
-class TestTextBlockRendering(SimplelayoutTestCase):
+class TestContentPageShowTitleBehavior(SimplelayoutTestCase):
 
     layer = FTW_SIMPLELAYOUT_FUNCTIONAL_TESTING
 
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,7 +5,8 @@ Changelog
 1.8.3 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Protect new show/hide title behavior.
+  [mathias.leimgruber]
 
 
 1.8.2 (2016-09-14)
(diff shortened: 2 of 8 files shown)


ftw/simplelayout/tests/test_contentpage_behavior.py
```diff
@@ -41,3 +42,13 @@ def test_title_shows_if_show_title_is_true(self, browser):
         self.assertEqual(
             ['A page'],
             browser.css('.documentFirstHeading').text)
+
+    @browsing
+    def test_show_hide_title_field_is_only_for_admins(self, browser):
```
jone: `only_for_admins` may be misleading since this is a side-effect here:
In this test it is "only for admins" because we have no active workflow and therefore the `rolemap.xml` counts, which does not grant contributors / editors.

But in production we will have a workflow, usually with `ftw.lawgiver`. In the `lawgiver.zcml` you are mapping the permission to the `edit` action group, therefore the standard behavior will be that all editors can hide the title.

Therefore I think the test name may be misleading.
How can we address that?
Should we just add a docstring to the test describing this issue?

Or is the `rolemap.xml` wrong because it does not go along with the `edit` action group? Then we should probably add editors / contributors in the `rolemap.xml` and then reconfigure the permission here so that we are explicitly building the situation in the test.

@maethu what do you think?

maethu: @jone hmm I guess it would be much simpler to simply ignore the permission with lawgiver... this way the rolemap.xml wins and the test is right. 

Currently I'm unable to say how we will use this feature in the future. The only usecase so far is to only permit Site Admins to edit this field.

jone: Ok. I'm not sure about what the best default is.

As an integrator I'd probably expect that the feature is available for all editing users, not just admins.

From the lawgiver perspective I think the action group `manage content settings` would best match.

How about:
1. Move the permission to `manage content settings`
2. No matter what's in the `rolemap.xml`, override the permission with `self.portal.manage_permission` in the test so that we have an isolated test scenario which does not take presumptions about how the feature should be used.

maethu: 👍  I changed that

maethu:
@jone updated.

PR #387, opened 2016-09-23
Drop support for plone standard theme (sunburst)

bierik:
The standard plonetheme sunburst is no longer supported. We recommend using a ftw.theming based theme.


bierik:
@jone updated

PR #388, opened 2016-09-26
Open external links in new tab.

shylux:
This PR implements the `external_links_open_new_window` option of plone.
This option marks links which point to other domains with `target="_blank"` to open them in a new tab.


maethu:
@bierik Is this the right place for this Plone specific feature?

ftw/simplelayout/browser/client/web/js/app/simplelayout/Block.js
```diff
@@ -49,6 +49,13 @@ define([
       return this;
     };
 
+    this.markLinks = function() {
+      var url = window.location.protocol + "//" + window.location.host;
+      // all http links (without the link-plain class), not within this site
+      $("a[href^=\"http\"]:not(.link-plain):not([href^=\"" + url + "\"])", this.element).attr("target", "_blank");
```
jone: Whether links should open in a new window or not is a configuration option in Plone; this change does _always_ open links in new windows, no matter what is configured in the property.

I would prefer if simplelayout does not _implement_ this behavior but only _trigger_ the Plone functions, which know and respect the property. See `mark_special_links.js`.

maethu: Or at least it should behave the same way!

shylux: Uh, but it does respect the property in the plone integration: https://github.com/4teamwork/ftw.simplelayout/pull/388/files#diff-b9536bb9b91bb6a804502b377ca00adcR206

ftw/simplelayout/browser/client/dist/simplelayout.js
```diff
@@ -587,7 +587,7 @@ define('app/simplelayout/Element',["app/simplelayout/idHelper", "jsrender"], fun
 define('app/simplelayout/Block',[
       "app/simplelayout/EventEmitter",
       "app/simplelayout/transactional",
-      "app/simplelayout/Element"
+      "app/simplelayout/Element",
```
shylux: Forgot to rebuild dev. Fixed it.

ftw/simplelayout/browser/client/web/js/app/simplelayout/Block.js
jone: Fair enough, didn't see that.

Did you think about just calling Plone's function? What was your conclusion?
Something like:
```javascript
if (typeof "scanforlinks" === "function") {
    scanforlinks();
}
```

jone: Should we only handle links within simplelayout containers?

shylux: The `mark_special_links.js` script does more than just setting the `target="_blank"`. It wraps links in `span`. It is not designed to be run multiple times.
I don't think we use those link wrappers anywhere. It looks like this was used for styling. So it is a bit outdated as we can use advanced css nowadays.

shylux: Yes. This is `ftw.simplelayout` after all. 😉  The links outside the simplelayout containers are already handled by `mark_special_links.js` (also configured by `external_links_open_new_window`).
This PR fixes the marking for simplelayout containers because they are added/edited dynamically.

jone: 👍 That's fine for me.

jone: @shylux but now the jquery selector selects all links on the page, right? Should it be limited to "the current context"?

shylux: @jone Nope it only selects links in `this.element`. I think the element in not even in the dom at this point.
http://api.jquery.com/jQuery/#jQuery-selector-context

jone: 🙈 👍

ftw/simplelayout/browser/client/dist/simplelayout.js
```diff
@@ -635,6 +635,13 @@ define('app/simplelayout/Block',[
       return this;
     };
 
+    this.markLinks = function() {
```
jone: As @maethu asked: is `simplelayout.js` the right place for this function?
It seems to me that `main.js` would be more apropriate.
@shylux has this impact on the test ability? What do you think about that?

shylux: The `simplelayout.js` is a generated file. The original function is in `Block.js`.
@bierik recommended this structure. I think it makes sense. `Block.js` for the implementation and `main.js` for the integration.
It is also way easier to test this way.

jone: Ok. I wouldn't have put it into the core block functionality, but I'm ok with keeping it there.
@maethu is it ok for you?

maethu: All other plone integration function are in the main.js?
We wait for @bierik ...

bierik: As @shylux sait the `main.js` ist just for integration and the code in `Block.js` is testable as well.
The block knows how to mark its content through the `this.element` reference.
So I would leave the `markLinks` function here.

PR #391, opened 2016-10-13
Auto resize youtube-videoblocks

raphael-s:
This makes videoblocks containing youtube video automatically resize to the available space given by their parent element without changing the width/height ratio.
This way there are no longer any black bars on the side of the block when the available space is bigger than the video itself.


bierik:
😍 thank you

PR #392, opened 2016-10-19
 Implement a inner_edit.json view, which allows us to edit an item in a folderish block.

maethu:
For example edit a file directly in a FilelistingBlock.
This view works almost like the block edit view, the difference is, that
the inner_edit.json view returns the rendered html of its parent (the Block).

FYI:
I'm gonna use this feature in the ftw.sliderblock first.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -264,11 +264,12 @@ Global Simplelayout configuration
 Create new Block representations
 --------------------------------
 
-Directly edit items in a folderish Block
+Directly edit items in a folderish block
 ----------------------------------------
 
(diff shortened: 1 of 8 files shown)


maethu:
aww hold on, it's failing

maethu:
Now it's working 😄 

🚬 🚬 🚬 🚬 🚬

maethu:
cc @bierik

bierik:
🚬 🚬 🚬 🚬 🚬 🚬 🚬  👍

Issue #396, opened 2016-11-07
Cannot delete layout without reload after removing last block


jone:
For deleting a layout the user has to move all blocks out of this layout.
But when doing so, the trash-can icon does not appear.
I browser reload is necessary for making the trash-can icon appear.
![design](https://cloud.githubusercontent.com/assets/7469/20050610/94431aa2-a4ca-11e6-923a-2117ecfa9c71.gif)

PR #398, opened 2016-11-10
Fix layout delete button

bierik:
The `inline` style on the toolbar is no longer valid.
Use `block` for display instead.

Closes https://github.com/4teamwork/ftw.simplelayout/issues/396

simplelayout_layout_delete-button


bierik:
@jone updated

PR #400, opened 2016-11-10
Don't allow all non-folderish types to be selectable as default page.

shylux:
Without this types like the TextBlock can be selected and would break the site.


maethu:
@jone  I added one commit to follow the best practices defined by plone for skins layers.

PR #401, opened 2016-11-16
🚧 Introduce showroom for galleryblock

bierik:
⚠️  Depends on https://github.com/4teamwork/ftw.showroom/pull/43

# Landscape

## Desktop

screen shot 2016-11-16 at 17 03 49
screen shot 2016-11-16 at 17 03 56

## Mobile

screen shot 2016-11-16 at 17 04 04
screen shot 2016-11-16 at 17 04 12

# Portrait

## Desktop

screen shot 2016-11-16 at 17 02 37
screen shot 2016-11-16 at 17 02 48

## Mobile

screen shot 2016-11-16 at 17 03 06
screen shot 2016-11-16 at 17 02 58
screen shot 2016-11-16 at 17 03 37


jone:
The close-button looks missplaced, giving the overlay an unfinished appearance.
Can we move the button to the corner of the overlay?
c8e32d1e-ac1e-11e6-9be5-95106444e68a

I also think that it would look better when we'd have rounded corners of the white area.

bierik:
`ftw.showroom` is not yet able to handle multiple galleries. This was not a requirement so far.

jone:
@bierik I think the Generic Setup profile of `ftw.showroom` is not installed properly.
I'm missing a dependency in `metadata.xml` and a `self.setup_install_profile('profile-ftw.showroom:default')` in the upgrade step.

bierik:
@jone updated

bierik:
@elioschmutz https://github.com/4teamwork/ftw.simplelayout/pull/401/files#diff-3f0aacf06a481c72c3187294eaad99b0R10 This should install the dependency doesn't it

elioschmutz:
ah yes! thats true. I was looking for a `metadata.xml` in the upgradestep directory

maethu:
I'm not yet ready to merge this, sry. I need some time.

jone:
In most projects, we are currently using colorbox for everything. Although it is not great, it is consistent.
When we start replacing it, we need to go all the way long and replace colorbox completely in order to have a consistent appearance again.
Introducing a new component costs a lot, since we may need to change styling in each customer project and we need to explain to each customer that and why there this is replaced, screenshots in documentations need to be replaced, etc.
This is currently not our priority.

Because of this impact we have decided that we wait for this issue to be more urgent and for the colorbox to mature in other projects such as GEVER.
We are not merging yet, but keeping the pull request open, so that we can continue here later.

bierik:
Because we are not able to update every legacy colorbox implementation these changes are not possible. So I will close this.

PR #403, opened 2016-11-25
Disable inline validation of the Plone forms in the overlay.

mbaechtold:
The main reason for this is the Plone file widget which causes some browsers to trigger the validation when the user wants to replace the file. The validation is triggered when the user clicks the button to select a file from the filesystem and he has to click it a second time for the file selector to open.

screenrecording

### DIFF
--- a/ftw/simplelayout/browser/resources/main.js
+++ b/ftw/simplelayout/browser/resources/main.js
@@ -2,6 +2,11 @@
 
   "use strict";
 
+  // Disable inline validation of the Plone forms in the overlay.
+  $(document).on('onLoad OverlayContentReloaded', '.overlay', function(){
+    $('.z3cformInlineValidation').removeClass('z3cformInlineValidation');
+  });
+
   function markLinks(element) {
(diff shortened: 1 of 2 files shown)


ftw/simplelayout/browser/resources/main.js
```diff
@@ -2,6 +2,14 @@
 
   "use strict";
 
+  // Disable inline validation of the Plone forms in the overlay.
+  $(document).on('onLoad', '.overlay', function(){
```
maethu: `$(document).on('onLoad', '.overlay', function(){` -> `$(document).on('onLoad OverlayContentReloaded', '.overlay', function(){`

ftw/simplelayout/browser/resources/main.js
```diff
@@ -2,6 +2,14 @@
...
+    $('.z3cformInlineValidation').removeClass('z3cformInlineValidation');
+  });
+  $(document).on('OverlayContentReloaded', '.overlay', function(){
```
maethu: ✂️

PR #409, opened 2016-12-20
Move option to hide a block to its own behavior.

mbaechtold:
This way the option can easily be added to other simplelayout block types, like the addressblock from "ftw.addressblock".

### DIFF
--- a/ftw/simplelayout/locales/de/LC_MESSAGES/ftw.simplelayout.po
+++ b/ftw/simplelayout/locales/de/LC_MESSAGES/ftw.simplelayout.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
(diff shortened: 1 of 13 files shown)


ftw/simplelayout/locales/de/LC_MESSAGES/ftw.simplelayout.po
```diff
@@ -11,15 +11,19 @@ msgstr ""
 "Plural-Forms: nplurals=1; plural=0\n"
 "Preferred-Encodings: utf-8 latin1\n"
 
+#: ./ftw/simplelayout/behaviors.zcml:37
+msgid "Adds an option to (visually) hide a block"
+msgstr "Fügt eine Option hinzu, die es erlabut, Blöcke visuell zu verstecken"
```
jone: `erlabut`

mbaechtold: Ist angepasst.

PR #412, opened 2017-01-09
Implement a simplelayout renderer object

maethu:
Implement a simplelayout renderer object, which allows us to render simplelayoutish structures without the simplelayout provider expression.

Further it's now possible to render only one layout of slot.

Since the diff is a little bit messed up, I'll explain what I've done. 
1. Basically I moved the code to from the `BaseSimplelayoutExpression ` to the `SimplelayoutRenderer`
2. The `BaseSimplelayoutExpression` still does the same, but uses the new `SimplelayoutRenderer`
3. The `SimplelayoutRenderer` has two new methods. A) Render the whole storage (e.g. default). B) Render one layout by storage and index.


ftw/simplelayout/browser/provider.py
```diff
@@ -33,37 +33,40 @@ def _prepare_html(self, text):
...
+        return self.slot_template()
 
-    def rows(self):
-        """ Return datastructure for rendering blocks.
-        """
-        page_conf = IPageConfiguration(self.context)
-        blocks = self._blocks()
+    def render_layout(self, index=None):
+
+        self.rows = self.storage.get(self.name, self._one_layout_one_column())
```
jone: I think this should be a local variable:
```diff
-         self.rows = self.storage.get(self.name, self._one_layout_one_column())
+         rows = self.storage.get(self.name, self._one_layout_one_column())
```

- it is only every used within `render_layout`
- it may be modified within `render_layout`, depending on `index`, and thus should not be exposed as its state is a side effect.

maethu: @jone 
it's used in the template... `view/rows`. Should implement a getter?

jone: Ah, ok 👍

PR #427, opened 2017-02-09
Prevent plone ajax spinner from being shown.

bierik:
The current approach to get rid of the spinner seems not to work properly.
Trying to unbind the `ajaxStart` event does not work in all cases.

So according to https://api.jquery.com/ajaxStart/,
> If $.ajax() or $.ajaxSetup() is called with the global option set to false, the .ajaxStart() method will not fire.

setting the `global` option to `false` never triggers the `ajaxStart` event.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,8 @@ Changelog
 1.17.1 (unreleased)
 -------------------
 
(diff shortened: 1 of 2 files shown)


mbaechtold:
Thank you very much!

PR #432, opened 2017-03-17
Fix toolbar delimiter styling.

bierik:
See https://github.com/4teamwork/edubsch.policy/issues/15

All the delimiters were visible even when the corresponding
action is disabled. So bind the delimiter to the action by using
a border instead of an additional delimiter DOM node.

screen shot 2017-03-17 at 14 27 23

### DIFF
--- a/ftw/simplelayout/browser/resources/ftw.simplelayout.js
+++ b/ftw/simplelayout/browser/resources/ftw.simplelayout.js
@@ -607,7 +607,7 @@ function Simplelayout(options) {
     },
     update: function update(event, ui) {
       if (ui.item.parent()[0] === this && !ui.sender) {
-        _EventEmitter2.default.trigger("layoutMoved", [ui.item.data().object]);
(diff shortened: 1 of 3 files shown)


jone:
@bierik how did it look before?
From the description it is unclear for me what exactly the effect of the change is.
Before / after screenshots would be helpful 😏

shylux:
@jone There was an artefact to the right of the layout. When only one layout exists in the slot you can neither move nor delete the layout which results in an empty action list looking like this: https://cloud.githubusercontent.com/assets/1375745/22730894/10893bb6-ede9-11e6-89f8-b85e0727031f.png

bierik:
@shylux Ja die Änderung ist gewollt:
Ich habe im https://github.com/4teamwork/ftw.simplelayout/pull/432/files#diff-6a359b8e11d4c593e9c2229333752a69L24 das Template für die Toolbox angepasst. Beim Kochen wird dann das `simplelayout.js` angepasst.

PR #434, opened 2017-03-22
Support Plone ordered container for sl blocks.

jone:
By synchronizing the simplelayout block order to Plone's ordered container, the catalog can sort blocks by their simplelayout order when using the objPositionInParent sort index.


ftw/simplelayout/configuration.py
```diff
@@ -180,6 +183,31 @@ def check_permission(self, new_state):
...
+    def get_ordered_blocks(self):
+        """Returns all blocks in the page, ordered by the order stored
+        in the page state.
+        """
+        uid_order = flattened_block_uids(self.load())
+
+        def block_sort_key(block):
+            uid = IUUID(block)
+            if uid not in uid_order:
+                return '~'
```
maethu: 😆

PR #435, opened 2017-03-22
Improve anchor extraction.

mbaechtold:
No longer rely on TinyMCE for the anchor extraction because it contains a bug when an empty value is searched for anchors (see https://github.com/plone/Products.TinyMCE/pull/153).

### DIFF
--- a/ftw/simplelayout/browser/anchors.py
+++ b/ftw/simplelayout/browser/anchors.py
@@ -1,7 +1,14 @@
-from ftw.simplelayout.interfaces import ISimplelayoutBlock
-from Products.Five.browser import BrowserView
 from Products.TinyMCE.browser.interfaces.anchors import IAnchorView
+from ftw.simplelayout.interfaces import ISimplelayoutBlock
+from lxml.html import fromstring
+from plone.app.textfield import IRichText
+from plone.dexterity.utils import iterSchemata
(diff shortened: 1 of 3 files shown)


ftw/simplelayout/browser/anchors.py
```diff
@@ -12,7 +19,36 @@ def listAnchorNames(self, *args, **kwargs):
...
+    def get_html_fields(self, obj):
+        fields = []
+        for schema in iterSchemata(obj):
+            for name, field in getFieldsInOrder(schema):
+                # Only consider RichText fields.
+                if IRichText.providedBy(field):
+                    fields.append(field)
+
+        # Filter empty fields (i.e. not having a value).
+        fields = filter(None, fields)
```
jone: I think this does not filter "empty fields" as promised, but rather fields that do not exist 🤔 
I would either move the value check to the `listAnchorNames` method or change this method to return values rather than fields.

Is it easily possible to add a test where the field value is actually `None`, so that we could detect that it crashed in this state?

mbaechtold: You are absolutely right. I have improved the code and added a test using an empty text block: https://github.com/4teamwork/ftw.simplelayout/pull/435/files#diff-96374b799c0ac7345fe9ba54e3d3f757R34

PR #436, opened 2017-03-23
Remove "addable_block_types".

jone:
The method `addable_block_types` on the simplelayout view is not in use and is therefore removed.

---

As I am customizing the `simplelayout-view` in my package I wanted to switch from `allowed_attributes` to `allowed_interface` so that I can inherit the configuration and do not need to update on new attributes in the future.

But I then came to the impression that the method `addable_block_types` was not used at all; the toolbar seems to work fine without.

@maethu please verify and confirm that this method is actually not used before merging this pull request!

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -7,6 +7,8 @@ Changelog
 
 - Make anchor extraction more robust. [mbaechtold]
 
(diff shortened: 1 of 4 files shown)


jone:
@maethu updated

PR #441, opened 2017-04-05
Disable kss spinner.

bierik:
Setting ajaxSetup.global to `false` is disabling every ajax event.
As described here https://api.jquery.com/ajaxSend/ the `ajaxSend` event is canceled as well.
So the request tracker was not working since https://github.com/4teamwork/ftw.simplelayout/pull/427 because all ajax events has been canceled.

First step is to remove the ajaxSetup.global setting to `false`.

Second step is to unbind the ajax event listeners from the kss.js
Two things are important here:
1. The RequestTracker.js has to come after the kss so that the listener is unbind properly
2. The unbind function has to be wrapped in a jQuery.ready callback because the kss.js is doing so as well

Third step is to bind our own ajax events to track all requests relevant for showing the spinner.

### DIFF
--- a/ftw/simplelayout/browser/resources/RequestTracker.js
+++ b/ftw/simplelayout/browser/resources/RequestTracker.js
@@ -97,11 +97,19 @@
     /*
       Disable kss spinner. In order for this to work, this file must be evaluated after kss-bbb.js.
       As you can see in the lib profile's jsregistry.xml.
-    */
-    $.ajaxSetup({ global: false });
 
-    $document.on("ajaxSend", function(event, jqxhr, params) { track(params.url); });
-    $document.on("ajaxComplete", function(event, jqxhr, params) { untrack(params.url); });
+      The `ajaxStart` and `ajaxStop` event is used in kss-bbb.js to toggle the spinner.
+      In simplelayout we have our own spinner so by unbinding these two events the spinner
+      is no longer shown.
+      So for every XHR request that has been started. it'll be tracked by the RequestTracker to
+      handle the spinner.
+    */
+    $(function() {
+      $document.unbind("ajaxStart");
+      $document.unbind("ajaxStop");
+      $document.on("ajaxSend", function(event, jqxhr, params) { track(params.url); });
+      $document.on("ajaxComplete", function(event, jqxhr, params) { untrack(params.url); });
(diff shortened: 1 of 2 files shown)


ftw/simplelayout/browser/resources/RequestTracker.js
```diff
@@ -98,10 +98,12 @@
...
-    $.ajaxSetup({ global: false });
-
-    $document.on("ajaxSend", function(event, jqxhr, params) { track(params.url); });
-    $document.on("ajaxComplete", function(event, jqxhr, params) { untrack(params.url); });
+    $(function() {
+      $document.unbind("ajaxStart");
+      $document.unbind("ajaxStop");
+      $document.on("ajaxSend", function(event, jqxhr, params) { track(params.url); });
+      $document.on("ajaxComplete", function(event, jqxhr, params) { untrack(params.url); });
+    })
```
jone: If I understand this correctly: you are defining a new function which is not referenced in a variable nor called after definition. When do you expect this code to be executed?

ftw/simplelayout/browser/resources/RequestTracker.js
```diff
@@ -98,10 +98,12 @@
       Disable kss spinner. In order for this to work, this file must be evaluated after kss-bbb.js.
       As you can see in the lib profile's jsregistry.xml.
     */
-    $.ajaxSetup({ global: false });
-
-    $document.on("ajaxSend", function(event, jqxhr, params) { track(params.url); });
-    $document.on("ajaxComplete", function(event, jqxhr, params) { untrack(params.url); });
+    $(function() {
+      $document.unbind("ajaxStart");
```
jone: I find it not obvious why you are unbinding events here. Could you add a short comment saying that you are unbinding KSS events?

bierik:
@jone updated

jone:
🤔

ftw/simplelayout/browser/resources/RequestTracker.js
jone: oh, didn't realize that this is a jquery ready callback 🙈

bierik: 🤔

PR #445, opened 2017-04-24
Simplelayout upgrade improvements

shylux:
Those changes were made during the (still ongoing) edubs migration.

This PR adds handler for `two column` and `two column with top row` layouts. It will generate an appropriate new simplelayout style layout for each and move the blocks based on the Slot marker interfaces.

### DIFF
--- a/ftw/simplelayout/migration.py
+++ b/ftw/simplelayout/migration.py
@@ -111,13 +111,13 @@ def move_sl_block_into_slot(old_page, new_page, block, slot_name):
     if slot_name == 'default' and page_state == page_configuration._default_page_config():
 
         # setup different layouts
-        if slot_name == 'default' and ISimplelayoutTwoColumnView.providedBy(old_page):
+        if ISimplelayoutTwoColumnView.providedBy(old_page):
             # two columns
             page_state[slot_name] = [
                 {'cols': [ {'blocks': []}, {'blocks': []} ]}
             ]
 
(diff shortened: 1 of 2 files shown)


ftw/simplelayout/migration.py
```diff
@@ -96,15 +103,61 @@ def __call__(self):
...
-    page_configuration = IPageConfiguration(page)
+def move_sl_block_into_slot(old_page, new_page, block, slot_name):
+    page_configuration = IPageConfiguration(new_page)
     page_state = page_configuration.load()
+
+    # initiate layout if it is in its initial state
+    if slot_name == 'default' and page_state == page_configuration._default_page_config():
+
+        # setup different layouts
+        if slot_name == 'default' and ISimplelayoutTwoColumnView.providedBy(old_page):
```
jone: The condition `slot_name == 'default'` is always `True` at this point and could be removed for simplification.

ftw/simplelayout/migration.py
```diff
@@ -96,15 +103,61 @@ def __call__(self):
...
+            # two columns
+            page_state[slot_name] = [
+                {'cols': [ {'blocks': []}, {'blocks': []} ]}
+            ]
+
+        elif slot_name == 'default' and ISimplelayoutTwoColumnOneOnTopView.providedBy(old_page):
```
jone: The condition `slot_name == 'default'` is always `True` at this point and could be removed for simplification.

PR #447, opened 2017-05-17
Add more sorting options to the gallery block.

mbaechtold:
The new sorting options are position in parent, id and portal type. Just like the file listing block. In fact, the same vocabulary holding the sorting options is used for the file listing block and the gallery block now.

### DIFF
--- a/ftw/simplelayout/tests/test_galleryblock.py
+++ b/ftw/simplelayout/tests/test_galleryblock.py
@@ -4,6 +4,7 @@ from ftw.builder import create
 from ftw.simplelayout.testing import FTW_SIMPLELAYOUT_CONTENT_TESTING
 from ftw.testbrowser import browsing
 from Products.CMFCore.utils import getToolByName
+from Products.CMFPlone.utils import parent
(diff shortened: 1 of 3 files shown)


ftw/simplelayout/tests/test_galleryblock.py
```diff
@@ -235,6 +236,29 @@ def test_hidden_galleryblock_not_visible_without_edit_permission(self,
...
+        block too.
+        """
+        gallery_block = create(Builder('sl galleryblock')
+                               .titled('My galleryblock')
+                               .within(self.page))
+        browser.login()
+        browser.visit(gallery_block, view='edit')
+        self.assertEqual(
+            [
+                ('portal_type', 'portal_type'),
```
maethu: @mbaechtold Can we remove portal_type? I mean it's always an image?

mbaechtold: Sure, I've updated the code.

mbaechtold:
Ping @maethu

PR #448, opened 2017-06-15
Change display mode of link tag around images in textblocks.

mbaechtold:
This makes the link span around the entire image. Until now it was not large enough.


maethu:
ohhh... this may have an impact on other customers, right?

mbaechtold:
Thanks for the hint, @maethu. I will move this change to the customer policy instead.

PR #450, opened 2017-06-30
Use a static vocabulary for the sort options of the gallery block.

mbaechtold:
Until now the sort options were based on the columns of the file listing block. If the colums were customized, unwanted sort option would have been applied to the gallery block.


mbaechtold:
Ping @maethu

PR #451, opened 2017-07-05
Improve gallery block while rendering truncated images.

mbaechtold:
Until now, if one of the contents of the gallery block contained a truncated error, the whole block could not be rendered. Render a fallback image in this case so the block is rendered and the broken image can be identified and be replaced by the content editors.

screenshot


maethu:
👍 I'll rebase and merge.

PR #453, opened 2017-07-17
Add anchor to new created blocks.

bierik:
The anchors are not handled by the block content.
So the anchor was just visible after a page refresh.

### DIFF
--- a/ftw/simplelayout/browser/resources/ftw.simplelayout.js
+++ b/ftw/simplelayout/browser/resources/ftw.simplelayout.js
@@ -105,8 +105,8 @@ function Block(content, type) {
 
   this.create({ type: type, content: content });
 
-  this.content = function (toReplace) {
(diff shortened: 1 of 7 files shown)


ftw/simplelayout/browser/resources/ftw.simplelayout.js
```diff
@@ -9319,7 +9323,7 @@ define(function (require, exports, module) {
 
 },{"amdefine":13}],58:[function(require,module,exports){
 /*!
- * EventEmitter v5.1.0 - git.io/ee
+ * EventEmitter v5.2.2 - git.io/ee
```
mbaechtold: How is the version number related to this pull request? I don't see any changes to the code of `EventEmitter` below...

bierik:
@mbaechtold Fixed pinning for the eventemitter so we always get the right version.

PR #428, opened 2017-02-09
Introduce configurable layout actions

bierik:

PR #456, opened 2017-08-02
Wrap columns with container.

bierik:
The change introduces an additional wrapper `div` with class `sl-columns`. 
This significantly improves the styling possibilities. 

For example:
- It's possible to define a background-color for all columns
- It's possible to manipulate the floating grid

This change may break stylings of ftw.simplelyout < 1.19.x
Also #428 introduced a layout wrapper. 

Example integration:
https://github.com/4teamwork/stadtwerk.web/pull/75

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,10 +5,12 @@ Changelog
 1.19.0 (unreleased)
 -------------------
 
-- Introduce asymmetric layouts. [Kevin Bieri, mathias.leimgruber]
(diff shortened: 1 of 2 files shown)


jone:
Please add some context and background. What is the motivation of this change? What is the impact? 
I'd like to know what the change is without looking at the code 🙄

maethu:
Changelog.

maethu:
Somehow the CI hook was not triggered (was probably down) I ran the tests locally and everything was fine.

PR #462, opened 2017-11-21
Synchronize page config with blocks plone.restapi integration

Nachtalb:
This implements a plone.restapi endpoint, with which you can synchronize the pages config with the existing blocks. This is needed after adding new blocks to a sl content page. Otherwise the blocks are added but are not visible in the FrontEnd.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -275,6 +275,15 @@ has the css class ``inneredit`` and a data attribute named ``uid`` containing th
 
 After editing the content, the view automatically reloads the block.
 
+Additional plone.restapi methods
+--------------------------------
+
+After creating blocks in a simplelayout content page they should be synchronized with the pages config. Otherwise
+the order in the frontend might me wrong. It also removes objects which are in the pages config but not in the page itself.
+
+To do this, you can simply send a RestAPI Post (more information about
+`plone.restapi <https://github.com/plone/plone.restapi>`_ ) request to the path of your page, appended with
+``@sl-synchronize-page-config-with-blocks``. A dict with ``added`` and ``removed`` block UIDs is returned.
 
 Run custom JS code
 ==================
--- a/ftw/simplelayout/restapi/configure.zcml
+++ b/ftw/simplelayout/restapi/configure.zcml
@@ -1,0 +1,15 @@
+<configure
+    xmlns="http://namespaces.zope.org/zope"
+    xmlns:plone="http://namespaces.plone.org/plone"
+    xmlns:zcml="http://namespaces.zope.org/zcml">
+
+  <include package="plone.rest" file="meta.zcml" />
+
+  <plone:service
+      method="POST"
+      name="@sl-synchronize-page-config-with-blocks"
+      for="ftw.simplelayout.interfaces.ISimplelayout"
+      factory=".services.SynchronizePageConfigWithBlocks"
+      permission="cmf.ModifyPortalContent"
+      />
+</configure>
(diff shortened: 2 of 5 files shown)


Nachtalb:
@jone Can you help me with this. I did it as we discussed but it is not working properly. It gives returns my added blocks as `added` but nothing has changed in the frontend. 
Also is there a way to test the `plone.restapi` endpoints?

jone:
@Nachtalb did you ensure that it is not cached locally in the browser? Simplelayout has some caches, thus the counter you sometimes see as GET-param in the URL.

README.rst
```diff
@@ -271,6 +271,15 @@ has the css class ``inneredit`` and a data attribute named ``uid`` containing th
 
 After editing the content, the view automatically reloads the block.
 
+Additional plone.restapi methods
+--------------------------------
+
+After creating blocks in a simplelayout content page they have to be synchronized with the pages config. Otherwise
+they wont show up in the FrontEnd. It also removes objects which are in the pages config but not in the page itself.
```
jone: `FrontEnd` => `frontend` 😉

README.rst
```diff
@@ -271,6 +271,15 @@ has the css class ``inneredit`` and a data attribute named ``uid`` containing th
...
+
+To do this, you can simply send a POST request to the path of your page, appended with
```
jone: You could also refer to the plone.restapi, as it is a standardized restapi endpoint.

ftw/simplelayout/restapi/configure.zcml
```diff
@@ -0,0 +1,16 @@
+<configure
+    xmlns="http://namespaces.zope.org/zope"
+    xmlns:plone="http://namespaces.plone.org/plone"
+    xmlns:zcml="http://namespaces.zope.org/zcml"
+    i18n_domain="opengever.api">
```
jone: 😏

ftw/simplelayout/restapi/services.py
```diff
@@ -0,0 +1,7 @@
+from plone.restapi.services import Service
+from ftw.simplelayout.configuration import synchronize_page_config_with_blocks
+
+
+class SynchronizePageConfigWithBlocks(Service):
+    def reply(self):
+        return synchronize_page_config_with_blocks(self.context)
```
jone: Did you verify if you can return a python data structure or if you need to return a `json` string?

ftw/simplelayout/restapi/services.py
```diff
@@ -0,0 +1,7 @@
+from plone.restapi.services import Service
+from ftw.simplelayout.configuration import synchronize_page_config_with_blocks
+
+
+class SynchronizePageConfigWithBlocks(Service):
+    def reply(self):
```
jone: I'd add an empty line here as of pep8.

ftw/simplelayout/restapi/services.py
Nachtalb: Yes. It is possible to just return a dict, the plone.restapi dies the rest😏

Nachtalb:
@jone made the requested changes. The problem with the not filling the containers was bgbern.ng specific. Its default container was not `default` but `top`. So I integrated the functionality to set the container you want the blocks in.

jone:
We have realized that it is not necessary to use this endpoint when the `default` simplelayout slot is used: the blocks will be rendered in the `default` slot when they are not explicitly configured in the simplelayout state.

Nachtalb:
@jone As you said, the problem we had is, that the name of the default slot was wrong. 

Now @maethu and I have realized that we need to run `synchronize-page-config-with-blocks` to make sure the order of blocks is correct. Atm, when adding blocks, they are not added to the pageconfig, like this sl does not know that it should order them in the way we created them. With adding them to the pageconfig the order of the blocks inside the sl page will have the same order as they have in the page itself (the order you see under `/path/folder_contents`).

PR #463, opened 2017-12-05
Mark active server actions on blocks.

shylux:
screen shot 2017-12-05 at 10 12 06

The block config is now provided in the `data-config` attribute. The config consists of the dict loaded from the `IBlockConfiguration` adapter, extended with the view name taken from the `IBlockProperties` adapter.

The block reload view was changed to now provide a JSON with the above described information and the block html in the `obj_html` attribute.

To determine which action is "active" the algorithm takes the parameter which would be sent with the request to the server and compares those with the current block configuration. If all the parameter on the action match the current block config the action is considered currently active and the class `active` is put on the element.


ftw/simplelayout/js/src/simplelayout/Block.js
```diff
@@ -56,6 +56,20 @@ export default function Block(content, type) {
...
 
+  this.updateToolbar = function() {
+    // check if configuration on block indicates that button is active
+    $(".block-server-action", this.toolbar.element).toArray().forEach(function(e) {
+      var buttonParameter = $(e).data();
+      var blockConfig = this.data("config") || {};
+
+      var active = Object.keys(buttonParameter).every(function(propertyName) {
+        return (propertyName in blockConfig &&
+                buttonParameter[propertyName] == blockConfig[propertyName]);
```
bierik: Always use `===` for comparison in javascript. This prevents the engine from type converting.

shylux:
Closing this PR. The change is processed in https://github.com/4teamwork/ftw.simplelayout/pull/464.

PR #464, opened 2017-12-05
Make layouts mutually exclusive & mark active actions on layout toolbar

shylux:
layout

Changed the way how layouts (1-3, golden ratio) are saved. They are now all saved in the `layout` attribute - which brings the desired effect that only one can be active at any given time.

Also mark the active actions on the layout toolbar.

/cc @jone


shylux:
@maethu Previously there were 3 layout attributes `layout31` `layout121` and `layout112`. You were able to activate all of them at the same time. Obviously only one applied, but depending on what other options were selected you could click a layout and nothing happened (because another layout took precedence). This resulted in bug reports from multiple clients.
My solution is to store those three options in one attribute `layout`. This way a new layout will override the old one.
If you want to add new options like color you will have to store them in a different attribute than `layout`. Just like the `golden-ratio` is an attribute on its own.

shylux:
@bierik The code in `Block` and `Layout` are indeed very similar. I didn't put it in `Element` because `Element` is not only parent of `Block` and `Layout`, but also `Layoutmanager` and `Toolbar`. So OOP wise  the empty `attachToolbar` and the new `updateToolbar` are already strange. And there is still a small difference between the `Block` and `Layout` implementation which has to be customized.
I think we should discuss possible solutions in person.

maethu:
@shylux Ahh nice, than my Use-case is still working 👍

maethu:
@shylux What is the impact of this for already existing websites, which already uses this (old) feature?

shylux:
@maethu There is an upgrade-step that migrates the config.

shylux:
`The toolbar action is not marked initially when I add a block`

I added the block config to the newly created block. The config contains only the view name since, there is no default config functionality.

shylux:
@bierik Rewrote the button active check to be generic for `Block` and `Layout`. It is not implemented in `Element`.

jone:
Outdated.

PR #466, opened 2017-12-15
Store the zoom level and the chosen map layer per mapblock. 

maethu:
This PR introduces two new features:
- Store the zoom level per block
- Store the chosen maplayer per block.

Bugfix:
- No longer register maplayers multiple times.
- Delete mapblock JS error. 

Info:
The zoom level was kind of implemented before, but it did not work, this is why there is a hack implemented for this purpose. --> https://github.com/4teamwork/ftw.simplelayout/compare/mle-map-zoom-layer?expand=1#diff-145524684bb7a847c44b606522518698R52

As the comment suggests, I'm not able to fix this in a better way, since it should already work. 

In action:

map

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -8,6 +8,9 @@ Changelog
 - Content page builder only synchronises page config if the
   page contains blocks. [mbaechtold]
 
+- Implement persisten map layer and zoom value.
+  Both can be chosen in the edit widget and will be reused on the display widget. [mathias.leimgruber]
+
 
 1.19.0 (2017-11-01)
 -------------------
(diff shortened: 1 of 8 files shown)


docs/HISTORY.txt
```diff
@@ -5,7 +5,10 @@ Changelog
 1.19.1 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Implement more specific MapLayers factory to prevent adding the same layers multiple times. [mathias.leimgruber]
+
+- Implement persisten map layer and zoom value.
```
mbaechtold: persisten -> persistent

PR #468, opened 2018-01-24
Add support for "ftw.gopip".

mbaechtold:
Unfortunately, we cannot write a test for this.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,8 @@ Changelog
 1.19.1 (unreleased)
 -------------------
 
+- Add support for "ftw.gopip". [jone, mbaechtold]
+
 - Optionally render a title of the mapblock. [mbaechtold]
 
 - Content page builder only synchronises page config if the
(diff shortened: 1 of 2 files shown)


jone:
A little more infos about that:
The problem occurs when changing the block order on the site root of an installation with 
`ftw.simplelayout` and `ftw.gopip`.
The problem is that from this point on, the reordering in folder contents does not work.
The reason is that the ordering indexed in the catalog (`getObjPositionInParent`) is incorrect, which is detected by the script applying the order change.

Teting: `ftw.simplelayout` does not depend on `ftw.gopip`, so in our testing environment we have the standard Plone implementation of the ordering, which looks up the ordering from the actual objects when queried in the catalog. Thus it cannot diverge. Thus it is hard to test without installing `ftw.gopip`.
The solution is just firing official Plone events, therefore we've decided to not test this change.

mbaechtold:
Thank you for you exhaustive explanation, @jone!

PR #470, opened 2018-02-14
Move title and show_title field into the coordinates schemata on MapBlock.

maethu:
This results in having only 1 tab, instead of two.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,8 @@ Changelog
 1.20.1 (unreleased)
 -------------------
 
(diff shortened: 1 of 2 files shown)


docs/HISTORY.txt
```diff
@@ -5,6 +5,7 @@ Changelog
 1.20.1 (unreleased)
 -------------------
 
+- Move title and show_title field into the coordinates schemata on MapBlock. [mathias.leimgruber]
```
mbaechtold: I'd try to be consistent with the format of the changelog and add a empty line between the list items.

maethu: 👍  done

PR #471, opened 2018-02-16
Fix delete block action.

mbaechtold:
Using a relative link causes an error when deleting a block on a content item which is used as the default page of a folder because the delete view is operating on the folder, not on the default page. The block to be deleted can not be found inside the folder because it only exists within the item used as the default page.


mbaechtold:
Ping @buchi.

PR #478, opened 2018-07-05
Quick fix sl baseurl + Fix problem in FF

### LINKED ISSUE #477: Compatibility Issues with plone 4.3.17 and greater.
Plone 4.3.17 introduced the base-url body attribute. 
Which behaves differently from the `<base>` tag. 
Simplelayout should get rid of that

maethu:
This implements falling back on the base tag if there is one. 
This will work until plone 5, plone 5 no longer has a base-tag. This issue will be fixed within the plone 5.1.x branch (good solution)

PLUS:
Add missing event parameter. --> Fixes FF

Resolves #477

### DIFF
--- a/ftw/simplelayout/browser/resources/main.js
+++ b/ftw/simplelayout/browser/resources/main.js
@@ -235,7 +235,7 @@
       }
     });
 
-    $(global.document).on("click", ".sl-layout .reload", function() {
+    $(global.document).on("click", ".sl-layout .reload", function(event) {
       event.preventDefault();
       var action = $(this);
       var layout = action.parents(".sl-layout").data().object;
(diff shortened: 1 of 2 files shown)


ftw/simplelayout/browser/resources/main.js
```diff
@@ -70,7 +70,7 @@
 
     var target = $("body");
 
-    var baseUrl = $("body").data("base-url") ? $("body").data("base-url") + "/" : $("base").attr("href");
+    var baseUrl = $("body").data("base-url") && $("base").length === 0 ? $("body").data("base-url") + "/" : $("base").attr("href");
```
jone: @maethu If I understand that correctly you are reverting the priority here. I think you could revert the condition and it would be easier understandable:
from: a && !b ? a : b
to:  b ?  b : a
Isn't that the same / does what you want?

maethu: yep ;-) and done

maethu:
I'll go for it, thanks for the hint!

PR #479, opened 2018-07-12
Implement staging for simplelayout pages

jone:
This pull request implements a staging toolset for staging simplelayout content pages.

- This is not batteries included: the staging implementation needs to be integrated on project level, since it is usually hooked up with a workflow, which is implemented in integration code on project level.
- The working copy is created by copying the content page with all content-relevant children, but not recursively including content pages. This allows for staging of root-level pages with tons of subpages.
- The integration code can decide wether the working copy is created in-place or in a centralized folder.
- The integration code can decide wether it is possible to have multiple working copies concurrently - there is no code for handling conflicts though.
- The staging is implemented for simplelayout content pages as primary object; it therefore only supports Dexterity as primary objects. Children may be Archetypes.
- The most important types of data are support when applying the working copy to the baseline, which is:
  - field values for Dexterity and Archetypes
  - scales are purged
  - Simplelayout block state is updated
  - Simplelayout page state is updated (UUIDs; reordering blocks; changing layouts)
  - children may be added or removed
  - update references in rich text fields (UUIDs)
- Viewlets with messages are included, informing the user that the content is a baseline or a working copy.
- ``ftw.trash`` is optionally supported: when applying a working copy to the baseline, the working copy itself is permanently deleted but the blocks removed from the baseline are only trashed.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -285,6 +285,47 @@ To do this, you can simply send a RestAPI Post (more information about
 `plone.restapi <https://github.com/plone/plone.restapi>`_ ) request to the path of your page, appended with
 ``@sl-synchronize-page-config-with-blocks``. A dict with ``added`` and ``removed`` block UIDs is returned.
 
+
+Staging
+-------
+
+Simplelayout provides integration level tools for setting up a staging solution for content pages.
+An ``IStaging`` adapter provides the functionality for making working copies and applying the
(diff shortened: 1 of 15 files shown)


jone:
@maethu this is ready for review. Thanks!

jone:
@maethu I have fixed both issues from your review.

PR #481, opened 2018-08-16
Image limits

elioschmutz:
This PR implement soft and hard image limits.

🚧 STILL UNDER CONSTRUCTION 🚧


ftw/simplelayout/contenttypes/browser/templates/textblock.pt
```diff
@@ -15,6 +16,7 @@
   <tal:image tal:condition="image">
     <div tal:attributes="class image/wrapper_css_classes">
         <div class="imageContainer">
+            <div metal:use-macro="view/context/@@ftw.simplelayout.images.macros/low_image_quality_indicator"></div>
```
jone: Why do we need to have a macro? It does not seem to be reused.
Could we move the tags from the macro here and get rid of the macro?

ftw/simplelayout/contenttypes/contents/textblock.py
```diff
@@ -70,6 +76,18 @@ class ITextBlockSchema(form.Schema):
 alsoProvides(ITextBlockSchema, IFormFieldProvider)
 
 
+class TextBlockImageLimitValidator(ImageLimitValidator):
+    """
```
jone: Is a docstring definition required here? 🤔 I'd remove the empty string if it does not matter 😉

ftw/simplelayout/images/configure.zcml
```diff
@@ -0,0 +1,18 @@
+<configure
+    xmlns="http://namespaces.zope.org/zope"
+    xmlns:plone="http://namespaces.plone.org/plone"
```
jone: (not neccessary - could be removed - optional)

ftw/simplelayout/images/limits.py
```diff
@@ -0,0 +1,62 @@
...
+        limits = self.get_limits_for(limit_type, identifier)
+        if self._validate(width, limits.get('width', 0)) and \
+                self._validate(height, limits.get('height', 0)):
+            return True
+        return False
+
+    def get_limits_for(self, limit_type, identifier):
+        """Returns the defined limits for the given identifier:
+
+        i.e. {'width': 100, 'heigh': 150}
```
jone: heigh -> height

ftw/simplelayout/contenttypes/contents/textblock.py
```diff
@@ -3,19 +3,25 @@
...
 from plone.directives import form
 from plone.namedfile.field import NamedBlobImage
+from z3c.form import validator
 from zope import schema
 from zope.i18n import translate
 from zope.interface import alsoProvides
 from zope.interface import implements
 
 
+TEXTBLOCK_IMAGE_LIMIT_IDENTIFIER = 'textblock'
```
jone: Could you update me on your thoughts here? Did you consider using the portal_type? Why did you discard that?

ftw/simplelayout/locales/de/LC_MESSAGES/ftw.simplelayout.po
```diff
@@ -556,8 +575,38 @@ msgstr "Datei hochladen"
...
+
+#. Default: "low image quality"
+#: ./ftw/simplelayout/images/templates/macros.pt:5
+msgid "low_quality_image_label"
+msgstr "Bildauflösung ist nicht optimal"
+
+#. Default: "The image is not in optimal resolution. Make sure, your image has at least the following dimensions: ${limit_str}"
+#: ./ftw/simplelayout/images/validators.py:44
+msgid "soft_limit_not_satisfied"
+msgstr "Das Bild ist nicht in einer optimalen Auflösung vorhanden. Stelle sicher, dass das Bild mindestens in folgender Auflsösung vorhanden ist: ${limit_str}"
```
jone: In Standardmodule verwenden wir normalerweise die Sie-Form. Ich würde das aus Konsistenzgründen hier auch machen.

ftw/simplelayout/images/validators.py
```diff
@@ -0,0 +1,95 @@
...
+        self._validate_hard_limit(value)
+        self._validate_soft_limit(value)
+
+    def _validate_hard_limit(self, value):
+        if not self._validate_limit_for('hard', value):
+            raise Invalid(self._image_hard_limit_invalid_error_message(value))
+
+    def _validate_soft_limit(self, value):
+        if not self._validate_limit_for('soft', value):
+            api.portal.show_message(
```
jone: We also have the static message on the image of the textblock. When does this limit occur? Dies it conflict with the message in the macro?

ftw/simplelayout/contenttypes/browser/templates/textblock.pt
elioschmutz: We'll use this markup for the galleryblock too... in the future, we'll use it for further implementations.

Do you think, it's useless?

ftw/simplelayout/contenttypes/contents/textblock.py
elioschmutz: Do you mean:

`TEXTBLOCK_IMAGE_LIMIT_IDENTIFIER = 'ftw.simplelayout.TextBlock'`?

That would be ok.

My first implementation was more generic. It was possible to set the config identifier like this:

`ftw.simplelayout.contenttypes.contents.textblock.ITextBlock.image`

and it was not necessary to define your own identifier. And why this long, complicated identifier? The Add-Form only provides the schema interface and a schema can have multiple image fields. 

But I didn't like it because it would have to implement a lot of adapters to make it more or less configurable if you want to customize it. Also I had to implement a lot of ugly lookups (i.e. the schema field from a contexts schema-interface and so on (I still have this branch if you are interested to see the previous implementation). Even the functions were lot complicated because this dynamic identifier.

The new implementation with the static identifier is much easier to understand, to configure and to customize.

ftw/simplelayout/images/validators.py
elioschmutz: The soft-limit portal-message will be visible for contenttypes not loaded within an overlay.

Currently, this is not the case. The message will not be shown anywhere. The Block-Implementation discards the portalmessages after creating the block. But I saw, that editing the block will not discard it. So I have to fix this anyway.

Currently, the `lowImageQualityIndicator`macro only shows the text "low image quality". But it would be helpful, if the user knows, what dimensions his image should be to match the requirements. So I could move the warning-message to the `lowImageQualityIndicator`:

Before:
bildschirmfoto 2018-08-23 um 09 44 42

After:
bildschirmfoto 2018-08-23 um 09 41 09

Another approach is to show the required dimensions within a tooltip after hovering the `lowImageQualityIndicator`

What is your opinion?

ftw/simplelayout/images/templates/macros.pt
```diff
@@ -0,0 +1,6 @@
+<div class="lowImageQualityIndicator"
+     metal:define-macro="low_image_quality_indicator"
+     tal:condition="showLowImageQualityIndicator"
```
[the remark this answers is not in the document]
elioschmutz: I thought implementing the whole logic within the macro is the way to do. If you want to use the macro, you just need to provide the required variables and that's it...

Is this a bad think?

Do you prefer to move the `tal:condition` outside of the macro?

elioschmutz:
@jone 👍thanks! I answered your questions/feedback.

ftw/simplelayout/images/templates/macros.pt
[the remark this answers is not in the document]
jone: Yes, I would move the `tal:condition` to the `textblock.pt` where you have `<div metal:use-macro`. The reason for moving it is to separate concerns -> that the macro should not have to rely on things defined outside.

ftw/simplelayout/contenttypes/contents/textblock.py
elioschmutz: I just run into problems with my implementation.

Using the portal-type would be the best. Within the add-form I could use the portal-type hardcoded. Or do you have a hint on how to properly get the portal-type within the validator?

If I do not need to differ the limits for different image-fields, this would be ⭐️ (and I think there was no such requirement 😉 )

jone: Yes, I would also go for the portal_type. Because it is simpler to understand and it should be good enough. We could also go with the fieldname `image` hard-coded (not sure about graphic block). So 👍

ftw/simplelayout/images/templates/macros.pt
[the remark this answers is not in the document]
elioschmutz: @jone I just tried to move the condition out of the macro. Unfortunately, this does not work.

```html
<div tal:condition="showLowImageQualityIndicator"
        metal:use-macro="view/context/@@ftw.simplelayout.images.macros/low_image_quality_indicator">
  Lorem ipsum
</div>
```

the macro seems to resolve before the `tal:condition`. This means, I would need to wrap it into a tales-block:

```html
<tal:block condition="showLowImageQualityIndicator"
    <div metal:use-macro="view/context/@@ftw.simplelayout.images.macros/low_image_quality_indicator">
  Lorem ipsum
     </div>
</tal:block>
```

Should I still move the condition outside of the macro?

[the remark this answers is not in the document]
jone: Yes, I'd still move it for separating concerns.

elioschmutz:
@jone 👍 i updated the PR. Could you take another look?

I also did the following changes for the soft-limit:

Before:
bildschirmfoto 2018-08-23 um 09 44 42

After:
bildschirmfoto 2018-08-23 um 09 41 09

To this, I have to access the context within the macro and I need to define an additional function to generate the message. You'll see the implementation in my last commit.
I don't see another way to do this.

PR #484, opened 2018-08-31
Use .eslintrc for linting

elioschmutz:
⚠️ MERGE #483 FIRST ⚠️ 

This PR introduces the .eslintrc for consistent linting.

It updates all the necessary files based on the .eslintrc-file with the JSPrettier (https://prettier.io/).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+root = true
+
+[*.js]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = false
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,230 @@
 {
+  "env": {
+    "browser": true,
+    "node": true,
+    "es6": true,
+    "mocha": true,
+  },
+  "globals": {
+    "nconf": true,
+    "assert": true,
+    "expect": true,
+    "fixture": true,
+  },
+  "parser": "babel-eslint",
+  "plugins": [],
+
   "parserOptions": {
     "ecmaVersion": 6,
-    "sourceType": "module"
+    "sourceType": "module",
+    "ecmaFeatures": {
+        "arrowFunctions": true,
+        "binaryLiterals": true,
+        "blockBindings": true,
+        "classes": true,
+        "defaultParams": true,
+        "destructuring": true,
+        "forOf": true,
+        "generators": true,
+        "modules": true,
+        "objectLiteralComputedProperties": true,
+        "objectLiteralDuplicateProperties": true,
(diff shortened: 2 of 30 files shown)


.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+[*.js]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = false
```
bierik: According to https://editorconfig.org/
you have to download a plugin for Atom, Emacs, VSCode and Sublime to make this editorconfig working. I do not think this is possible to maintain this plugin across all the developers.
The eslinter itself should be able to conform to each of its formats.

.eslintrc
```diff
@@ -1,6 +1,230 @@
...
+        "objectLiteralShorthandProperties": true,
+        "octalLiterals": true,
+        "regexUFlag": true,
+        "regexYFlag": true,
+        "spread": true,
+        "superInFunctions": true,
+        "templateStrings": true,
+        "unicodeCodePointEscapes": true,
+        "globalReturn": true,
+        "jsx": true
```
bierik: We use no `jsx` in this project.

.eslintrc
```diff
@@ -1,6 +1,230 @@
...
+    // The following rules are included for compatibility with JSHint and JSLint.
+    // While the names of the rules may not match up with the JSHint/JSLint counterpart,
+    // the functionality is the same.
+    //
+    "max-depth": [0, 3], // specify the maximum depth that blocks can be nested (off by default)
+    "max-len": [2, 100, 2], // specify the maximum length of a line in your program (off by default)
+    "max-params": [2, 5], // limits the number of parameters that can be used in the function declaration. (off by default)
+    "max-statements": 0, // specify the maximum number of statement allowed in a function (off by default)
+    "no-bitwise": 0, // disallow use of bitwise operators (off by default)
+    "no-plusplus": 2, // disallow use of unary operators, ++ and -- (off by default)
```
bierik: Where did you get those rules from?
For better readability I would recommend using the https://www.npmjs.com/package/eslint-config-airbnb-base ruleset from eslint. You can simple extend those rules https://eslint.org/docs/user-guide/configuring#extending-configuration-files and you don't have to define them separately.

.eslintrc
```diff
@@ -1,6 +1,230 @@
...
+    "sourceType": "module",
+    "ecmaFeatures": {
+        "arrowFunctions": true,
+        "binaryLiterals": true,
+        "blockBindings": true,
+        "classes": true,
+        "defaultParams": true,
+        "destructuring": true,
+        "forOf": true,
+        "generators": true,
```
bierik: Generators are not fully supported in babel-preset-es-2015. If you want to use generators we should update the babel ecma version.

PR #490, opened 2018-10-08
Add support for www.youtube-nocookie.com video urls.

maethu:
- More DSGVO compatible (only 5 cookies :-D).
- As iframe - no youtplayer since that one loads a like 20 cookies.


ftw/simplelayout/contenttypes/contents/videoblock.py
```diff
@@ -25,6 +25,13 @@ def is_youtube_url(url):
     return parsed_url.netloc == 'youtu.be' and len(path) == 2 and path[-1]
 
 
+def is_youtube_nocookie_url(url):
+    # ttps://www.youtube-nocookie.com/embed/UUrddqT9i_s
```
jone: h

PR #498, opened 2018-12-14
Fix ftw.trash dependency

busykoala:
ftw.trash was referenced even though it is not installed by default, resulting in a crash upon start.


busykoala:
👍

Issue #496, opened 2018-12-03
PAM and Staging Support doesn't work


maethu:
``` 
Module ZPublisher.Publish, line 138, in publish
Module ZPublisher.mapply, line 77, in mapply
Module ZPublisher.Publish, line 48, in call_object
Module ftw.lawgiver.browser.modifystatus, line 22, in __call__
Module bgbern.ng.browser.staging, line 16, in revise_baseline_in_working_copy
Module ftw.simplelayout.staging.staging, line 104, in create_working_copy
Module ftw.simplelayout.staging.staging, line 158, in _create_clone
Module OFS.CopySupport, line 234, in manage_pasteObjects
Module zope.event, line 31, in notify
Module zope.component.event, line 24, in dispatch
Module zope.component._api, line 136, in subscribers
Module zope.component.registry, line 321, in subscribers
Module zope.interface.adapter, line 585, in subscribers
Module zope.component.event, line 32, in objectEventNotify
Module zope.component._api, line 136, in subscribers
Module zope.component.registry, line 321, in subscribers
Module zope.interface.adapter, line 585, in subscribers
Module OFS.subscribers, line 132, in dispatchObjectCopiedEvent
Module zope.container.contained, line 153, in dispatchToSublocations
Module zope.component._api, line 136, in subscribers
Module zope.component.registry, line 321, in subscribers
Module zope.interface.adapter, line 585, in subscribers
Module plone.app.multilingual.subscriber, line 133, in createdEvent
AttributeError: REQUEST
```

Nachtalb:
Steps to reproduce: 
1. Create a sl content page 
2. Add a sl TextBlock to it
3. Trigger staging by workflow change (usually eg. to revise the page)
4. Error as shown above

jone:
@maethu I have analyzed the problem:
- https://github.com/4teamwork/ftw.simplelayout/pull/516/ adds a PR, testing that PAM and staging works together.
- I've realized that the problem is already fixed on PAM's `2.x` branch; I've made a request for a release https://github.com/plone/plone.app.multilingual/issues/335

PR #516, opened 2019-02-12
Test for plone.app.multilingual supported staging

jone:
Adds a test for proving that plone.app.multilingual is supported.

The currently released version `plone.app.multilingual = 2.0.3` does not work well with our staging; see #496.

The branch `2.x` works though, because the change https://github.com/plone/plone.app.multilingual/commit/2f4a8e95f4dd54ea5faa33879241fa1ec5a6808a fixes the issue.

### DIFF
--- a/ftw/simplelayout/testing.py
+++ b/ftw/simplelayout/testing.py
@@ -7,6 +7,7 @@ from ftw.testing.layer import ComponentRegistryLayer
 from path import Path
 from persistent.list import PersistentList
 from persistent.mapping import PersistentMapping
+from plone.app.multilingual.browser.setup import SetupMultilingualSite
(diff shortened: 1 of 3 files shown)


sources.cfg
```diff
@@ -6,3 +6,8 @@ extensions +=
     mr.developer
 
 auto-checkout =
+    plone.app.multilingual
+
+
+[branches]
+plone.app.multilingual = 2.x
```
jone: https://github.com/plone/plone.app.multilingual/issues/335

maethu:
thanks @jone I'm happy it was not our issue and it's already fixed!

maethu:
@Nachtalb  Please use the 2.x branch for your project, so you can go ahead

PR #526, opened 2019-05-10
Revert commit removing IReferenceable and add tests

busykoala:
The commit `abc8` mistakenly removed IReferenceable in plone4. In consequence references were not added to the reference catalog. This adds it again plus a test showing that it is working again.


busykoala:
> Could you add a changelog entry?
> What was the missing part that the tests now pass?

They do not pass yet. I try printing the html into the output now to find out.

busykoala:
The test mentioned won't pass because ftw.trash overwrites the deleting method. I removed this test because of that.

jone:
@busykoala you were missing the changelog entry, could you add it?

busykoala:
@jone I commited this just minutes after I mistakenly merged it without
https://github.com/4teamwork/ftw.simplelayout/pull/527.

PR #530, opened 2019-05-20
Allow map preferences for behaviours

Nachtalb:
This will be used for ftw.addressblock to work correctly.
The selector looks for an id starting with `field-widget` because otherwise, it would also select the widgets wrapper div. To get the zoomlevel / maplayer we filter for ending with. Together with the starting filter, we can get all zoomlevel / maplayer widgets from or not from a behaviour.

This is in conjunction with https://github.com/4teamwork/ftw.addressblock/pull/25

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 ------------------
 
 - Add macros and slots to block views for easy customization. [jone]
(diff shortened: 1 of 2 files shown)


Nachtalb:
Rebased onto latest master, waiting for tests before merging.

PR #535, opened 2019-06-12
Fix mapblock.js for plone 5

Nachtalb:
On plone 5 we don't have the same tab structure in forms as in plone 4 anymore. This lead the mapblock to not initialising the map when clicking on a tab. To fix this we use new selectors for plone 5 and fallback to plone 4 selectors if no tabs could be found.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,7 +5,7 @@ Changelog
 2.1.1 (unreleased)
 ------------------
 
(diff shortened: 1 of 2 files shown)


Nachtalb:
This error can be observed with the latest plone 5 branch version of the `ftw.addressblock` with the `geo` extra.  --> https://github.com/4teamwork/ftw.addressblock/pull/27

PR #539, opened 2019-06-27
Fix click event not caught on plone 5 due to preventation through plone.mockup

Nachtalb:
The JS which handles the tabs in edit forms and so on stops the
propagation of the click event. So as an alternative we use the `mouseup`
event, which is fired after the click and is not prevented by plone.

This is a part of https://github.com/4teamwork/ftw.simplelayout/pull/535


maethu:
I fixed that one in #564 by using the "click" event instead of "onClick"

PR #540, opened 2019-07-23
 Galleryblock: show image description as caption.

maethu:
- A new option is given to the user, whether the description of the image is used a caption in the overlay or not. 
- The option is enabled by default
- If there's no description and the option is enable it falls back to the old implementation (shows the alt text of the image as caption in the overlay.

Screen Shot 2019-07-23 at 17 24 55
Screen Shot 2019-07-23 at 17 24 43

### DIFF
--- a/ftw/simplelayout/locales/de/LC_MESSAGES/ftw.simplelayout.po
+++ b/ftw/simplelayout/locales/de/LC_MESSAGES/ftw.simplelayout.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
-"POT-Creation-Date: 2019-04-23 12:14+0000\n"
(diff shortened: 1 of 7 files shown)


ftw/simplelayout/locales/de/LC_MESSAGES/ftw.simplelayout.po
```diff
@@ -574,6 +574,11 @@ msgstr "25% - 75% Verteilung"
 msgid "label_layout_inverse"
 msgstr "Umkehren / Spiegeln"
 
+#. Default: "Show description of images as legend"
+#: ./ftw/simplelayout/contenttypes/contents/galleryblock.py:69
+msgid "label_legend"
+msgstr "Beschreibung von den Bilder als Legende in der Grossansicht verwenden"
```
jone: Ich würde 
`Beschreibung der Bilder`
statt
`Beschreibung von den Bilder`

PR #541, opened 2019-08-22
Preserve children's children order when applying working copies

Nachtalb:
The same approach is taken as in here:, by moving all items to top it updates and preserves the correct order.
https://github.com/4teamwork/ftw.simplelayout/blob/6781711d351714a0f7a9255c2a8a855ea7c466ab/ftw/simplelayout/configuration.py#L208-L212

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -13,6 +13,7 @@ Changelog
 
 - Fix contenttypes and mapblock js resources. [mathias.leimgruber]
 
+- Preserve childrens children order when applying working copies [Nachtalb]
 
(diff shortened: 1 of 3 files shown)


Nachtalb:
@jone I have added the test to check that the order of sub pages is not changed.
https://github.com/4teamwork/ftw.simplelayout/blob/8c5710dce0f9a6446e19c678112bf746205a25b4/ftw/simplelayout/tests/test_staging.py#L209-L214

Nachtalb:
@maethu I give this to you because @jone is absent atm.

PR #546, opened 2019-09-16
Install contenttypes extras by default and deprecate the extra itself

Nachtalb:
We do this for multiple reasons:
- To fix and prevent future dependency problems. At the moment
it was not the case that when you install ftw.simplelayout without the
[contentypes] extra the plone instance could not be installed due to
missing dependencies. To prevent this in the future and fix the current
bug we now install all the contenttypes dependencies by default.
- The [contenttypes] extra is used everywhere. We do not know of any
usage of simplelayout without the contenttypes extra. Simplelayout
without the contenttypes doesn't make much sense because the types added
is what makes simplelayout usable in the first case.

There is room for further improvements regarding the dependencies and the split between the core and the contenttypes. But those are not yet to be done. This PR is mainly to fix and prevent further dependency errors.


setup.py
```diff
@@ -81,21 +71,35 @@
       zip_safe=False,
 
       install_requires=[
-          'setuptools',
-
+          # 'cssutils', # TODO: is this used anywhere?
```
jone: ✂️

Nachtalb: Done via the GitHub interface, imma use this more now I like that you can simply make those changes 👍

Nachtalb:
Rebased onto latest master.

PR #549, opened 2019-10-29
Save file as blobs

Nachtalb:
Files should be saved as blobs and not in the database as binary data.
This also fixes a bug that with images that are rotated via
exif:Orientation. When saving images as NamedImage the rotation fix by
plone.namedfile was not applied correctly, when using NamedBlobImage
though this works like a charm and fixes the images rotation.


jone:
> Do we need to migrate existing data?

Discussed: only files uploaded with `ftw.simplelayout >= 2.0.0` are affected. These files are stored in the ZODB. A migration would simply move them outside of the ZODB as blobs.
As we have only very few test installations which are affected and a potential upgrade step would be very expensive (time consuming) we've decided that we are not providing a migration here.

Nachtalb:
Rebased onto latest master.

PR #554, opened 2019-10-31
Fix save translation to match the rest of plone in plone 4 / 5

Nachtalb:
In plone 4 the "Save" from `PloneMessageFactory` translates to "Sichern" whereas the rest of plone uses "Speichern" from `plone.dexterity.i18n`. So we use this as well to make sure the "Save" message is the same across the site and across plone versions.


Nachtalb:
Oops, yeah let me do that quickly...

Nachtalb:
Done

Nachtalb:
Rebased onto latest master.

PR #563, opened 2019-11-14
Add contenttypes to mapblock profile.

busykoala:
Mapblock is only used when contenttypes are installed also. Therefore it would be nice if there is a dependency so that when mapblock is getting installed also the contenttypes are getting installed.

This is particularly useful when we need to add mapblock as a dependency in another package where we otherwise had to put both instead of just one.

TODO:
- [x] Fix uninstall

```
AssertionError: u'Index: types/ftw.simplelayout.ContentPage.xml\n=============================== [truncated]... != ''
- Index: types/ftw.simplelayout.ContentPage.xml
- ===================================================================
- --- types/ftw.simplelayout.ContentPage.xml	2019/11/14 19:23:41.627561 GMT+1
- +++ types/ftw.simplelayout.ContentPage.xml	2019/11/14 19:23:41.627761 GMT+1
- @@ -19,6 +19,7 @@
-    <element value="ftw.simplelayout.ContentPage"/>
-    <element value="ftw.simplelayout.VideoBlock"/>
-    <element value="ftw.simplelayout.GalleryBlock"/>
- +  <element value="ftw.simplelayout.MapBlock"/>
-   </property>
-   <property name="allow_discussion">False</property>
-   <property name="default_view">@@simplelayout-view</property> : Quickinstaller seems not to uninstall everything.
```

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -9,6 +9,7 @@ Changelog
 
 - No longer accidentally load google maps api js twice. [mathias.leimgruber]
 
+- Add ftw.simplelayout.contenttypes profile when mapblock is added by default. [busykoala]
 
 
 2.4.1 (2019-11-14)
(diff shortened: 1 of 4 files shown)


jone:
Please rebase; the changelog is conflicting 😉

PR #568, opened 2019-11-27
Remove at builder.

busykoala:
🚧 Do not merge yet. 🚧 

Close https://github.com/4teamwork/ogb/issues/106

The dx builder for the link is already registered and conflicts with the older at builder. By removing the old builder this problem can be solved.


busykoala:
Close in favour of #567

PR #570, opened 2019-12-05
Unpin six.

busykoala:
Close https://github.com/4teamwork/ogb/issues/114

Pinning six does not seem to be necessary. Because it results in a version conflict I unpin it. The version conflict is a result of the release of ftw.testbrowser 2.0.0.

If files are not opened in binary mode testbrowser will fail here with the below error:
https://github.com/4teamwork/ftw.testbrowser/blob/2833d166e0a6b1f10c7d14da3f0560f4d15de868/ftw/testbrowser/form.py#L368
```
'unicode' does not have the buffer interface
```
Here first a broken example and then a working one:
```
file_ = open(Path('whatever.txt').abspath(), 'r')
browser.fill({u'File Fäldli': file_})
```
```
file_ = open(Path('whatever.txt').abspath(), 'rb')
browser.fill({u'File Fäldli': file_})
```

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -7,6 +7,7 @@ Changelog
 
 - Old versions of ftw.sliderblock depended on this product to require collective.upload, which
 is no longer done after version 2.0.  We raise ImportError on startup to make this clear. [djowett-ftw]
+- Fix tests not opening files in binary mode (necessary to work with the latest release of ftw.testbrowser. [busykoala]
 
 
 2.5.1 (2019-11-29)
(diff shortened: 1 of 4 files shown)


busykoala:
@buchi file mode leads to error in testbrowser 8a53f4b ;-)

docs/HISTORY.txt
```diff
@@ -7,6 +7,7 @@ Changelog
 
 - Old versions of ftw.sliderblock depended on this product to require collective.upload, which
 is no longer done after version 2.0.  We raise ImportError on startup to make this clear. [djowett-ftw]
+- Fix tests not opening files in binary mode (necessary to work with the latest release of ftw.testbrowser. [busykoala]
```
jone:
```suggestion
- Fix tests not opening files in binary mode (necessary to work with the latest release of ftw.testbrowser). [busykoala]
```

busykoala: Merci ;-)

PR #573, opened 2019-12-11
Provide precompiled bundle

buchi:
main.js doesn't work without require.js (Plone 4) yet


buchi:
Restored compatibility with Plone 4.

Currently the ftw.simplelayout module in `ftw/simplelayout/js/src/` has to be built separately as before with `yarn run grunt dist`.

`yarn build` only builds the integration in  `ftw/simplelayout/browser/resources/` which includes the built ftw.simplelayout module.

busykoala:
I don't exactly understand what the `private: true` in package.json does and if it is necessary. If it was, please add this as well...

jone:
> I don't exactly understand what the `private: true` in package.json does and if it is necessary. If it was, please add this as well...

See https://devdocs.4teamwork.ch/plone/javascript/ftw-resource-bundling-plone5/#bundling-with-yarn-and-friends-outside-plone

jone:
I'll make the upgrade step...

PR #574, opened 2019-12-16
Fix error when file is uploaded as ZPublisher.HTTPRequest.FileUpload

Nachtalb:


Nachtalb:
ATM there is a zcml conflict in the tests because the lastes released version of 4teamwork/ftw.file also adds this utility. This is removed and replaced with 4teamwork/ftw.uploadutility in this PR 4teamwork/ftw.file#99

jone:
@Nachtalb we should also use the `ftw.uploadutility` here.

jone:
@Nachtalb what is the state here?

Nachtalb:
Rebased onto latest master.
@jone it's ready now. `ftw.file` has been released and thus the tests do not fail anymore.

PR #583, opened 2020-01-06
Repair overlay styles

djowett-ftw:
Resolves https://github.com/4teamwork/izug.organisation/issues/1720

Looking somewhat better 

Screenshot 2020-01-06 at 15 50 27

Screenshot 2020-01-06 at 15 52 24

Screenshot 2020-01-06 at 15 57 47

Screenshot 2020-01-06 at 15 52 03

referencewidget could be clearer on styling expandable (traversable) items, but I guess that's a different issue

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,7 +5,7 @@ Changelog
 2.5.6 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Repair overlay styles. [djowett-ftw]
 
 
 2.5.5 (2019-12-20)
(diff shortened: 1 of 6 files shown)


djowett-ftw:
Checked this in Plone 4 with blueberry theme and seems ok, though the referencewidget overlay is the same size as the overlay behind it....


Screenshot 2020-01-06 at 16 48 54


Screenshot 2020-01-06 at 16 49 06


Screenshot 2020-01-06 at 16 49 22

djowett-ftw:
Without the blueberry theme (or ftw.theming) activated, this looks as follows - but we don't have a plain CSS in this product to fix it....

Screenshot 2020-01-06 at 16 26 36

djowett-ftw:
> Yay, cool 🎉
> 
> Could you make a changelog entry? 😉

Good point! and ... done :-)

PR #595, opened 2020-03-17
Fix youtube integration. The iframe is now stored in `f` not `a`.

maethu:
Apparently  we use a internal API of the youtube player to determine the ratio of the iframe.

This internal API has changed, probably yesterday (16.03.2020).

We are accessing a variable called `a`, which hold the iframe, but now it's stored in `f`.

Screen Shot 2020-03-17 at 19 16 25
Screen Shot 2020-03-17 at 18 27 20

There is some more noise regarding `postMessage` but this seems not be a problem right now.


maethu:
Please also check the backport https://github.com/4teamwork/ftw.simplelayout/pull/596

ftw/simplelayout/browser/resources/videoblock.js
```diff
@@ -12,7 +12,7 @@
   function resizeAll() { $.map($(".sl-youtube-video"), resizeYoutubePlayer); }
 
   function onPlayerReady(player) {
-    var iframe = player.target.a;
+    var iframe = player.target.f;
```
Nachtalb: This alone won't work. The bundle won't reload the file. We either have to create a second file where video block is compiled into or we have to load it into the default simplelayout bundle.

Nachtalb:
@maethu I've made a commit with the said changes.

ftw/simplelayout/browser/resources/videoblock.js
maethu: @Nachtalb I don't really get it. But I'm don't know anyway how to deal with that the right i generated a new simplelayout-contenttypes-compiled.js and it worked locally in dev and prod mode. But I think you guys know what you're doing.

ftw/simplelayout/contenttypes/upgrades/20200317204342_remove_obsolete_videoblock_js_bundle/upgrade.py
```diff
@@ -0,0 +1,9 @@
+from ftw.upgrade import UpgradeStep
+
+
+class RemoveObsoleteVideoblockJsBundle(UpgradeStep):
+    """Remove obsolete videoblock js bundle.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
```
maethu: @Nachtalb There should be a condition right? To only apply this change for plone5?

ftw/simplelayout/browser/resources/videoblock.js
Nachtalb: What do you mean you generated a new simplelayou-contenttypes-compiled.js? This file does not exist...

ftw/simplelayout/contenttypes/upgrades/20200317204342_remove_obsolete_videoblock_js_bundle/upgrade.py
Nachtalb: Oh yeah you're right..

maethu:
@Nachtalb Your commit did not work at all. since simplelayout.js is only loaded while logged in the videos are still not working on zg.ch 😞

maethu:
@Nachtalb This was obviously also the reason why it was in a different bundle.

PR #599, opened 2020-04-01
Always load videoblock js not only when logged in

### LINKED ISSUE #597: videoblock JS only loaded if logged in
With #595 the videoblock.js has been moved to the simplelayout bundle, which only get's served if you are logged in. 

videos should also be played if you are anonymous. 

Please add a second bundle for simplelayout, which is available also for anonymous user.
Note: I does not make sense to serve simplelayout in the default bundle, since 99.9%  would not be used.

Nachtalb:
resolves #597

This is needed for the videoblock to work correctly. Thus we create a
second js that should always be loaded.

Not logged in, js loaded, video works fine, music is great
---
image


djowett-ftw:
Did you mean to omit `merge_with`?

djowett-ftw:
@maethu (and @Nachtalb) I added merge_with => default and corrected the dependency declarations on the bundle.  Ok?

maethu:
> @maethu (and @Nachtalb) I added merge_with => default and corrected the dependency declarations on the bundle. Ok?

Yep, IMHO that's good!! Can you also rebase and create a new compiled.js?

djowett-ftw:
> @djowett-ftw Yep, IMHO that's good!! Can you also rebase and create a new compiled.js?

Done the rebase. `grunt build` doesn't change anything, which is reassuring! I also removed a line from Gruntfile.js that really doesn't need to be there (if someone did a `require('simplelayout_main')` we'd get 2 copies of that code in both default.js and logged-in.js IIUC)

PR #601, opened 2020-04-17
Fix galleryblock uploadzone

djowett-ftw:
Also depend on fixes in ftw.colorbox for getting overlay for galleryblocks


ftw/simplelayout/browser/resources/main.js
```diff
@@ -82,7 +82,7 @@
     });
   }
 
-  $(document).on('ready', function(){
+  $(document).ready(function(){
```
maethu, line 85: What's the difference here? Does this change the behavior?

djowett-ftw: It seems to be related to this comment in https://api.jquery.com/ready/ ....

      "There is also $(document).on( "ready", handler ), deprecated as of jQuery 1.8 and removed in jQuery 3.0. Note that if the DOM becomes ready before this event is attached, the handler will not be executed."

I suspect that's exactly what happened in our case (just due to the volume of JavaScript)

djowett-ftw: I tested this - and I _did_ see this making a difference

maethu: 👍

PR #606, opened 2020-05-20
Add AliasBlock

busykoala:
This PR adds a new sl-block (AliasBlock).

Jira Ref: https://4teamwork.atlassian.net/browse/OGB-35

TODO:
- [x] Tests
  - [x] Add AliasBlock
  - [x] Test logic & view
- [x] Add type (and some basic setup: resource folder, upgrade, ...)
- [x] Make use of https://github.com/4teamwork/ftw.autofeature
- [x] Add translations
- [x] Implement logic
- [x] Add styling (using blueberry)
- [x] Add changelog entry
- [x] Document (`README.rst`)

image

### DIFF
--- a/ftw/simplelayout/aliasblock/__init__.py
+++ b/ftw/simplelayout/aliasblock/__init__.py
@@ -1,0 +1,4 @@
+def initialize(context):
+    """Required for beeing able to uninstall the aliasblock product.
+    """
+    pass
--- a/ftw/simplelayout/aliasblock/browser/templates/aliasblock.pt
+++ b/ftw/simplelayout/aliasblock/browser/templates/aliasblock.pt
@@ -1,0 +1,3 @@
+<div:aliasblock class="sl-alias-block">
+  <h2 tal:content="view/block_title" tal:condition="view/block_title">Title</h2>
+</div:aliasblock>
--- a/ftw/simplelayout/aliasblock/browser/templates/aliasblock.pt
+++ b/ftw/simplelayout/aliasblock/browser/templates/aliasblock.pt
@@ -1,3 +1,22 @@
-<div:aliasblock class="sl-alias-block">
-  <h2 tal:content="view/block_title" tal:condition="view/block_title">Title</h2>
-</div:aliasblock>
+<html xmlns:tal="http://xml.zope.org/namespaces/tal"
+      xmlns:i18n="http://xml.zope.org/namespaces/i18n"
+      i18n:domain="ftw.simplelayout">
+
+<div class="sl-alias-block">
+
+  <tal:ref-valid condition="not: context/alias/isBroken">
+      <a tal:attributes="href context/alias/to_path"
+         class="sl-alias-block-visit-block"
+         i18n:translate="label_visit_block">
+        &#128279; Visit embedded block
+      </a>
+    <tal:alias-block content="structure view/get_referenced_block_content" />
+  </tal:ref-valid>
+
+  <tal:ref-invalid condition="context/alias/isBroken">
+    <p i18n:translate="label_invalid_reference">The embedded block does not exist.</p>
+  </tal:ref-invalid>
+
+</div>
+
+</html>
--- a/ftw/simplelayout/aliasblock/browser/templates/aliasblock.pt
+++ b/ftw/simplelayout/aliasblock/browser/templates/aliasblock.pt
@@ -5,12 +5,14 @@
 <div class="sl-alias-block">
 
   <tal:ref-valid condition="not: context/alias/isBroken">
+    <tal:authorized condition="view/get_referenced_block_content">
       <a tal:attributes="href context/alias/to_path"
          class="sl-alias-block-visit-block"
          i18n:translate="label_visit_block">
         &#128279; Visit embedded block
       </a>
-    <tal:alias-block content="structure view/get_referenced_block_content" />
+      <tal:alias-block content="structure view/get_referenced_block_content" />
+    </tal:authorized>
   </tal:ref-valid>
 
   <tal:ref-invalid condition="context/alias/isBroken">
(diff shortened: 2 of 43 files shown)


maethu:
> @busykoala Generally it's ok to have a new type and make it just _not_ addable by default. Since we have to do a per customer integration. So IMHO there's no need to have it's own profile

But since you already went down this path you can also keep it as separate profile 👍

maethu:
@busykoala Cool, I like it so far 💚

ftw/simplelayout/aliasblock/__init__.py
```diff
@@ -0,0 +1,4 @@
+def initialize(context):
```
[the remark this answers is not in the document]
maethu: @busykoala You sure this is required, see my comment above.

ftw/simplelayout/aliasblock/contents/aliasblock.py
```diff
@@ -0,0 +1,24 @@
...
+from zope import schema
+from zope.interface import alsoProvides
+from zope.interface import implements
+
+
+class IAliasBlockSchema(form.Schema):
+    """AliasBlock for simplelayout
+    """
+
+    title = schema.TextLine(
```
maethu, line 15: @busykoala Please also add the show_title bool field, so we have more consistent behavior for the different blocks.

ftw/simplelayout/configure.zcml
```diff
@@ -29,6 +29,7 @@
 
     <include package=".contenttypes" zcml:condition="have ftw.simplelayout:contenttypes" />
     <include package=".mapblock" zcml:condition="have ftw.simplelayout:contenttypes:mapblock" />
+    <include package=".aliasblock" zcml:condition="have ftw.simplelayout:contenttypes:mapblock" />
```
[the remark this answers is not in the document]
maethu: @busykoala 👀

ftw/simplelayout/aliasblock/browser/templates/aliasblock.pt
```diff
@@ -0,0 +1,3 @@
+<div:aliasblock class="sl-alias-block">
+  <h2 tal:content="view/block_title" tal:condition="view/block_title">Title</h2>
```
maethu, line 2: Add condition for the mentioned `show_title` field.

ftw/simplelayout/aliasblock/contents/aliasblock.py
busykoala: The field is removed in further implementation. There will be no title field as specified...

ftw/simplelayout/aliasblock/browser/templates/aliasblock.pt
busykoala: This is removed in further implementation. There is no title in this block as specified.

ftw/simplelayout/configure.zcml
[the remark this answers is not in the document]
busykoala: I fixed that in 37bb953b6408bb20fd6a1c0cda8d1c65eb47a59a. Thanks!

ftw/simplelayout/aliasblock/__init__.py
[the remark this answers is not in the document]
busykoala: This is required for the installation of the profile. The test will fail because the layer `ftw.testing.genericsetup.ZCMLLayer` can't be setup anymore. Also see the mapblock profile implementation as a reference.

ftw/simplelayout/aliasblock/contents/aliasblock.py
```diff
@@ -0,0 +1,57 @@
...
+from plone.dexterity.content import Item
+from plone.directives import form
+from plone.directives.form import widget
+from z3c.relationfield.schema import RelationChoice
+from zope.interface import alsoProvides
+from zope.interface import implements
+
+
+def get_alias_path(widget):
+    return '/'.join(api.portal.get().getPhysicalPath())
```
maethu, line 16: Why always start at the root?

ftw/simplelayout/aliasblock/hooks.py
```diff
@@ -0,0 +1,34 @@
+from Products.CMFCore.utils import getToolByName
+from plone.registry.interfaces import IRegistry
+from zope.component import getUtility
+import logging
+import pkg_resources
+import re
+
+
+IS_PLONE_5 = pkg_resources.get_distribution('Products.CMFPlone').version >= '5'
+PROFILE_ID = 'profile-ftw.book:default'
```
maethu, line 10: book? and PROFILE_ID is never used anyway

ftw/simplelayout/aliasblock/contents/aliasblock.py
busykoala: It doesn't say in the specifications and since it would be pointless to reference a block close to the page having the alias block I decided to just start at the root.
I guess that's something we need to configure for a customer if they have a certain preference and otherwise just go with something.

ftw/simplelayout/aliasblock/hooks.py
busykoala: Good catch!

busykoala: Fixed that and the unused imports in bd67805

ftw/simplelayout/aliasblock/contents/aliasblock.py
maethu: @busykoala There is a quite hidden feature https://github.com/4teamwork/ftw.referencewidget/blob/d24ee07563f7f1657299b02a6376ca3630e9bb4c/ftw/referencewidget/browser/utils.py#L89 I would probably use "parent" 😊

busykoala: Nice, thanks!

PR #607, opened 2020-05-27
Replace title with reference widget.

busykoala:
This PR is part of the new alias block implementation and adds the reference widget used for selecting the block whose content will, later on, be rendered within the alias block.

### DIFF
--- a/ftw/simplelayout/testing.py
+++ b/ftw/simplelayout/testing.py
@@ -17,7 +17,8 @@ from plone.testing import z2
 from plone.testing import zca
 from unittest import TestCase
 from zope.configuration import xmlconfig
-import ftw.simplelayout.tests.builders
+import ftw.referencewidget.tests.widgets  # noqa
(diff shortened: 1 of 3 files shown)


ftw/simplelayout/testing.py
```diff
@@ -79,7 +81,7 @@ class SimplelayoutTestCase(TestCase):
     def assert_recursive_persistence(self, structure):
 
         def is_persistent(item):
-            if not isinstance(item, basestring) and not isinstance(item, int):
+            if not isinstance(item, basestring) and not isinstance(item, int):  # noqa
```
maethu, line 84: Why the # noqa? we usually allow 120 letters per line.

busykoala: `basestring` can't be found according to flake8.

maethu: ahh you use flake8 of a python 3 😏

PR #612, opened 2020-06-02
Fix path issue on startpages (non sl pages using sl appearance).

busykoala:
The problem was that when non sl page were using the sl appearance the context on e.g. a startpage (plone site) would be the application instead of the platform. This resulted in an issue because the block object could not be found because no site is set on this context.

Adding the whole path instead of the relative part resolves this issue and the image size can be switched now.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,7 +5,7 @@ Changelog
 2.6.2 (unreleased)
 ------------------
 
-- Nothing changed yet.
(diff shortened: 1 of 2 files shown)


busykoala:
This issue can be solved adding a base tag to a custom design like https://github.com/4teamwork/plonetheme.blueberry/pull/160.

This is the preferred way because otherwise many ftw packages would need this change in their code (not only sl).

PR #614, opened 2020-06-04
Do not allow to add a reference to a contentpage which contains a aliasblock

maethu:
This prevents possible recursions and weird behavior for the editor.

I also adde more understandable error messages then just "Constraint not satisfied".

### DIFF
--- a/ftw/simplelayout/aliasblock/contents/aliasblock.py
+++ b/ftw/simplelayout/aliasblock/contents/aliasblock.py
@@ -17,8 +17,15 @@ class AliasBlockSelectable(DefaultSelectable):
 
     def is_selectable(self):
         selectable = super(AliasBlockSelectable, self).is_selectable()
-        return selectable
+        is_sl_page = self.content.portal_type == 'ftw.simplelayout.ContentPage'
 
+        # Don't allow sl pages containing another AliasBlock
+        if is_sl_page:
+            return not bool(filter(
+                lambda item: item.portal_type == 'ftw.simplelayout.AliasBlock',
+                self.content.objectValues()
+            ))
+        return selectable
 
 
 def get_selectable_blocks():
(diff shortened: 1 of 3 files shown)


ftw/simplelayout/aliasblock/contents/aliasblock.py
```diff
@@ -17,8 +17,15 @@ class AliasBlockSelectable(DefaultSelectable):
 
     def is_selectable(self):
         selectable = super(AliasBlockSelectable, self).is_selectable()
-        return selectable
+        is_sl_page = self.content.portal_type == 'ftw.simplelayout.ContentPage'
```
Nachtalb, line 22: Wouldn't it make more sense to check if the content provides `IContentPage` to make it possible to use objects which inherit from `ftw.simplelayout.ContentPage` without problems?

ftw/simplelayout/aliasblock/contents/aliasblock.py
```diff
@@ -17,8 +17,15 @@ class AliasBlockSelectable(DefaultSelectable):
...
 
+        # Don't allow sl pages containing another AliasBlock
+        if is_sl_page:
+            return not bool(filter(
+                lambda item: item.portal_type == 'ftw.simplelayout.AliasBlock',
```
Nachtalb, line 27: Here again, why not use the interface instead of the name, in case some integration needs an object inheriting from the alias block.

ftw/simplelayout/aliasblock/contents/aliasblock.py
maethu: @Nachtalb As of right now we only support exact this type. Other similar types like NewsFolder, or News may not work or may not make sense, since we don't show the archive portlet for example.

ftw/simplelayout/aliasblock/contents/aliasblock.py
maethu: There will be no other AliasBlock. If you wanna change the behavior of the AliasBlock, you can change the `Scheme` and `Class` in the FTI, but it should not change the name.

PR #615, opened 2020-06-04
Implement rendering of sl page in a alias block.

maethu:
Looks not too bad otherwise 😊 

Screen Shot 2020-06-04 at 17 03 20

Screen Shot 2020-06-04 at 17 03 30

- I added a `assert_html` helper in order to have a `udiff` output if it differs.

### DIFF
--- a/ftw/simplelayout/tests/test_alias_block.py
+++ b/ftw/simplelayout/tests/test_alias_block.py
@@ -11,6 +11,7 @@ from unittest import TestCase
 from z3c.relationfield import RelationValue
 from zope.component import getUtility
 from zope.intid.interfaces import IIntIds
+import difflib
 import transaction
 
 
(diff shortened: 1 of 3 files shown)


ftw/simplelayout/tests/test_alias_block.py
```diff
@@ -27,6 +28,17 @@ def setUp(self):
 
         self.intids = getUtility(IIntIds)
 
+    def assert_html(self, page_html, aliasblock_html):
+        diff = difflib.unified_diff(
```
Nachtalb, line 32: According to https://docs.python.org/2.7/library/difflib.html#difflib.unified_diff, this function is for creating a diff between two lists. It seems as if you give it full strings. Wouldn't this then diff every single character to one another, meaning if one word is already different the diff would get quite big?

maethu: Good point I'll have a look, I did work on my end when I changed something in the middle. But indeed the api suggest something else 🤔

Nachtalb: If it is a problem I think you could just split the string by newlines and it would be ok.

maethu: @Nachtalb That was a really good call. I only tested some white spaces... and they worked. 

Before:
Screen Shot 2020-06-09 at 11 41 07

After:
Screen Shot 2020-06-09 at 11 43 35

Nachtalb: 🎉

PR #616, opened 2020-06-05
Show a nicer message if the content is not accessable.

maethu:
Otherwise the "The block could not be rendered" shows to a author.

### DIFF
--- a/ftw/simplelayout/aliasblock/browser/templates/aliasblock.pt
+++ b/ftw/simplelayout/aliasblock/browser/templates/aliasblock.pt
@@ -5,7 +5,7 @@
 <div class="sl-alias-block">
 
   <tal:ref-valid condition="not: context/alias/isBroken">
-    <tal:authorized condition="view/get_referenced_block_content">
+    <tal:authorized condition="view/has_view_permission">
       <a tal:attributes="href context/alias/to_path"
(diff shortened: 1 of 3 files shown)


ftw/simplelayout/aliasblock/browser/templates/aliasblock.pt
```diff
@@ -15,6 +15,10 @@
     </tal:authorized>
   </tal:ref-valid>
 
+  <tal:not-authorized condition="not: view/has_view_permission">
+    <p i18n:translate="label_invalid_permission">The content is no longer accessible for you</p>
```
Nachtalb, line 19: Update locales or do we want one big locales updated in the end (either way is ok we just mustn't forget about it)?

maethu: I'll update at the end, once everything is approved. Updating locales on the way is a pain in the ass.

djowett-ftw, line 19:
```suggestion
    <p i18n:translate="label_invalid_permission">The content is no longer accessible to you</p>
```

maethu: @djowett-ftw changed

PR #620, opened 2020-06-08
Add google map js to aliasblock add/edit form. In case the block returns map content this is necessary.

maethu:
This fixes the fowling JS error:

Screen Shot 2020-06-08 at 09 17 26

### DIFF
--- a/ftw/simplelayout/aliasblock/browser/viewlet.py
+++ b/ftw/simplelayout/aliasblock/browser/viewlet.py
@@ -1,0 +1,23 @@
+from ftw.simplelayout.aliasblock.contents.interfaces import IAliasBlock
+from plone import api
+from plone.app.layout.viewlets.common import ViewletBase
+from Products.Five.browser.pagetemplatefile import ViewPageTemplateFile
+from zope.component import queryMultiAdapter
+
+
+def geo_settings_installed():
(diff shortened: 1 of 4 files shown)


ftw/simplelayout/aliasblock/browser/viewlet.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+
+class AliasBlockFormViewlet(ViewletBase):
+    index = ViewPageTemplateFile('templates/aliasblock_form_viewlet.pt')
+
+    def update(self):
+        super(AliasBlockFormViewlet, self).update()
+
+        is_aliasblock = IAliasBlock.providedBy(self.context)
```
djowett-ftw, line 20: Is it not possible to use `for="ftw.simplelayout.contenttypes.contents.interfaces.IAliasBlock"` in the ZCML?

maethu: @djowett-ftw It would be, but only on the edit_form, the add form is called on the container, therefore on the simplelayout page. This way only 1 viewlet needs to be registered instead of twice almost the same.

djowett-ftw: ok, yeah, makes sense. thanks!

PR #621, opened 2020-06-08
Update readme and changelog.

maethu:


README.rst
```diff
@@ -156,6 +156,11 @@ provided to jump to the original content.
 The selectable blocks are returned by
 ``ftw.simplelayout.aliasblock.contents.aliasblock.get_selectable_blocks``.
 
+It's also possible to select a whole ContentPage to render within the AliasBlock.
+It's not possible to select another ContentPage which already contains an AliasBlock.
+Alle Blocks within this package are supported (incl. MapBlock). ftw.sliderblock and ftw.addressblock are working too.
+Other Blocks may work, but they are not tested.
```
Nachtalb, line 162: I'd make it a bit shorter and remove unnecessary information.

```suggestion
It's also possible to select a whole ContentPage to render within the AliasBlock unless that page already contains another AliasBlock (prevent recursion).
All blocks within this package are supported (incl. MapBlock).
Blocks from other packages may work but are not tested.
```

djowett-ftw, line 162: As commented elsewhere - Blocks from other packages won't work because the list of aliasable blocks is hard coded

Nachtalb: cood catch 👍

maethu:
@djowett-ftw  @Nachtalb updated.

PR #622, opened 2020-06-09
Update translations.

maethu:
3 new msgs


ftw/simplelayout/locales/ftw.simplelayout.pot
```diff
@@ -402,6 +402,16 @@ msgstr ""
 msgid "description_teaser_fieldset"
 msgstr ""
 
+#. Default: "The selected Content cannot be selected"
```
djowett-ftw, line 405: I'd lowercase `Content` for the English :-)

maethu: I'll change that

PR #624, opened 2020-06-10
Fix AliasBlock text for link to referenced ContentPage

Nachtalb:
When we reference a page we shouldn't tell the user that it is a block. 
image


Nachtalb:
Rebased onto latest master.

PR #631, opened 2020-06-18
Add ftw.logo support for plone root opengraph integration.

maethu:
- ftw.simplelayout has now a soft dep on ftw.logo.
- This fixes the opengraph integration if both are installed. Before og:image returned the hard coded logo.jpg.


Nachtalb:
Nice!

PR #634, opened 2020-07-01
Update aliasblock integration JS for full multi columns/layouts contentpage support.

maethu:
Those were findings from the Stadtwerk integration. 

The issue was if the referenced contentpage had multiple layouts with multiple columns it might had side effects on reloading a layout (for example changing the ratio from 50/50 to 25/75).


ftw/simplelayout/js/src/simplelayout/Layoutmanager.js
```diff
@@ -81,7 +81,8 @@ export default function Layoutmanager() {
     this.commit();
     $(".sl-layout", restoreElement).each(function() {
       if ($(this).closest('.sl-alias-block').length === 0) {
-        self.insertLayout().restore(this, self, $(".sl-column", this).length);
+        const cols = $(".sl-column", this).filter(function(index, element){ return $(element).closest('.sl-alias-block').length === 0 })
```
djowett-ftw, line 84: missing ; ?

maethu:
> I don't see how it will impact the page but the code looks clean.

It prevents managing columns and blocks within a aliasblock :-)

PR #647, opened 2020-08-13
Escape special chars for opengraph title and description.

maethu:
Since we don't use TAL to render the content some signs need to be escaped.

Example if somebody used a double quote in the description:
Screen Shot 2020-08-13 at 15 52 42

I adjusted the test data. The assertions do not change since the test browser unescapes those signs again.
But the test would fail with the new data and without the changes.

### DIFF
--- a/ftw/simplelayout/opengraph/og_site_root.py
+++ b/ftw/simplelayout/opengraph/og_site_root.py
@@ -42,9 +42,24 @@ class PloneRootOpenGraph(object):
         return registry.forInterface(
             ISimplelayoutDefaultSettings, check=False)
 
+    def _escape(self, text):
+        """More or less taken from https://wiki.python.org/moin/EscapingHtml
+        """
+        html_escape_table = OrderedDict([
+            ("&", "&amp;"),
+            ('"', "&quot;"),
+            ("'", "&apos;"),
+            (">", "&gt;"),
+            ("<", "&lt;"),
+        ])
+
+        for sign in html_escape_table.items():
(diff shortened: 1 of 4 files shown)


ftw/simplelayout/opengraph/og_site_root.py
```diff
@@ -42,9 +42,24 @@ def _settings(self):
         return registry.forInterface(
             ISimplelayoutDefaultSettings, check=False)
 
+    def _escape(self, text):
+        """More or less taken from https://wiki.python.org/moin/EscapingHtml
+        """
+        html_escape_table = OrderedDict([
```
Nachtalb, line 48: If we already do it manually we could make it stronger by using pythons inbuilt modules:
```python
>>> import cgi
>>> data = u'<this is="ä">test'
>>> cgi.escape(data, quote=True).encode('ascii', 'xmlcharrefreplace')
'&lt;this is=&quot;&#228;&quot;&gt;test'
```

`cgi.escape` does the same as you are doing here and the `.encode(..)` additionally also makes sure that umlauts are escaped.

maethu: @Nachtalb it does not handle `&`. and this is one of the official proposed solution from the https://wiki.python.org/moin/EscapingHtml 

I considered using cgi modul to but then I still had to do the & manually. IMHO this one is the most explicit solution. 

Umlauts do not need to be encoded.

Nachtalb (approved):

PR #649, opened 2020-09-04
Fix link integrity check on Plone 5

Nachtalb:
In Plone 4 we get a dict. In Plone 5 we get a list of dicts because the API allows for checking multiple objects at once. https://4teamwork.atlassian.net/browse/OGB-624


maethu (changes requested):
Please add a changlog entry.

PR #652, opened 2020-09-08
Remove margin from sl images on small screens

djowett-ftw:
These images are 100% wide on small screens (see line 128 and following), so they don't need margin there.

This might seem wide-ranging, but it's the right place to clear it up IMO, 
when would we really want to have margins on 100% width images?

For https://4teamwork.atlassian.net/browse/OGB-672

before

Screenshot 2020-09-08 at 16 38 31

after

Screenshot 2020-09-08 at 16 37 37

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,7 +5,8 @@ Changelog
 2.7.12 (unreleased)
 -------------------
 
-- Nothing changed yet.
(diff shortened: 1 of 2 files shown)


djowett-ftw:
Also an issue on Dietikon.  

Screenshot 2020-09-08 at 16 57 43

maethu (approved):
Please add a changelog entry, thanks!

PR #653, opened 2020-09-10
Not necessary to add the plain IGeoreferenceable interface as a behavior 

djowett-ftw:
... if IBlockCoordinates is there

This is deprecated and produces the warning:
```
        plone.dexterity.schema No behavior registration found for behavior named: 'collective.geo.geographer.interfaces.IGeoreferenceable'
```
which got raised as something to be fixed in bl.web (see https://4teamwork.atlassian.net/browse/OGB-708)

 
IBlockCoordinates adds IGeoreferenceable as a marker interface anyway

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,7 +5,7 @@ Changelog
 2.7.13 (unreleased)
 -------------------
 
-- Nothing changed yet.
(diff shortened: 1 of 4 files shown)


maethu (approved):
Please add a changlog entry, otherwise 👍

maethu (approved):

PR #654, opened 2020-09-16
AliasBlock rendering improvements. 

maethu:
This basically avoids completely from having issues rendering the block. 

Either "he content is no longer accessible to you" or "The embedded block does not exist." will be displayed.
A failure while rendering the block does not happen anymore.

Before the code usually broke when it tried to access `isBroken` or `to_object` of None type value.


Nachtalb (approved):
Has this also been done to the 2.x versions or is this 1.x only?

maethu:
> Has this also been done to the 2.x versions or is this 1.x only?

I'll make a PR for the 2.x too.

PR #656, opened 2020-09-29
Fix loading jquery 2 times and with this creating conflicts

Nachtalb:
Without the depends on plone our JS was loaded before the
plone-compiled JS. This somehow resulted in require.js loading the
jQuery a second time even though jQuery was already present. Why exactly
this happens and why loading plone as the first script fixes this we
don't know - but it does.


maethu (approved):
💚 👐

PR #657, opened 2020-10-06
Move objects to the target if they are created within the working copy.

maethu:
This preserves the UUIDS and internal links created within the working
copy are still working.

The major change here is, that if the content gets newly created in the working copy. The UUID will be preserved, since the content gets moved the target page instead of copied over.

### DIFF
--- a/ftw/simplelayout/staging/staging.py
+++ b/ftw/simplelayout/staging/staging.py
@@ -293,6 +293,7 @@ class Staging(object):
         uuid_map = uuid_map or {}
         uuid_map[IUUID(source)] = IUUID(target)
         target_children_map = {IUUID(obj): obj for obj in self._get_children(target, condition)}
+        source_ids = source.objectIds()
         self._copy_field_values(source, target)
         self._purge_scales(target)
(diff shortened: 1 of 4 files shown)


ftw/simplelayout/staging/staging.py
```diff
@@ -303,7 +303,7 @@ def _apply_children(self, source, target, condition=None, uuid_map=None):
                 target_child = target_children_map.pop(target_uid)
                 self._apply_children(source_child, target_child, uuid_map=uuid_map)
             else:
-                target_child = self._copy_new_obj(source_child, target)
```
Nachtalb, line 306: This is the only place where we used `_copy_new_obj` so you can remove that method.

Nachtalb (changes requested):

ftw/simplelayout/staging/staging.py
maethu: 👍  I'll go after the tests :-( seems like I broke quite a bit

Nachtalb (approved):

PR #659, opened 2020-10-14
Update page config is recursively

Nachtalb:
Only updating the configuration of the directly copied page leads to broken page configurations and thus all blocks are thrown into the default slot which should not happen.


Nachtalb:
Rebased onto latest master.

maethu (approved):

PR #671, opened 2020-12-16
Add height field to videoblock in order to set a fixed height for vimeo/youtube nocookie block.

maethu:


Nachtalb (approved):

ftw/simplelayout/contenttypes/contents/videoblock.py
```diff
@@ -54,6 +54,10 @@ class IVideoBlockSchema(form.Schema):
                       u'VIDEO_ID'),
         required=True)
 
+    player_height = schema.TextLine(
```
Nachtalb, line 57: Maybe we should add a help message that we want a full height spec including `px` etc. Otherwise I think people will put in "400" and then ask why it won't work.

maethu: @Nachtalb if there is unit, the browser assume the value is in Pixel. So it does not matter if the put in 400, or 400px

Nachtalb: Oh didn't know that 👍

PR #676, opened 2021-03-01
Check all incoming references upon block deletion. This includes AliasBlocks

maethu:
Also add a better title for aliasBlocks.

The incoming ref change only works with Plone 5.
Since the link integrity check in plone 4 does not support DX relations (zc.relations)


maethu:
@Nachtalb FYI

